PR #38, opened 2015-01-14
modify gogo to generate (more) stable import names

PatrickDropbox:
This reduces the churn in the generated code (at dropbox, we check in the generated code into our repos)

### DIFF
--- a/protoc-gen-gogo/generator/helper.go
+++ b/protoc-gen-gogo/generator/helper.go
@@ -91,7 +91,7 @@ func (this *pluginImports) NewImport(pkg string) Single {
 func (this *pluginImports) GenerateImports(file *FileDescriptor) {
 	for _, s := range this.singles {
 		if s.IsUsed() {
-			this.generator.P(s.Generate())
+            this.generator.PrintImport(s.Name(), s.Location())
 		}
 	}
 }
@@ -99,7 +99,10 @@ func (this *pluginImports) GenerateImports(file *FileDescriptor) {
 type Single interface {
 	Use() string
 	IsUsed() bool
-	Generate() string
+Name() string
+  Location() string
+
+	Generate() string  // DEPRECATED
 }
 
 type importedPackage struct {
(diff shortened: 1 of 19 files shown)


awalterschulze:
We also check in generated code into our repos.

If I run make all
> panic: assignment to entry in nil map
Please run at least make all and make testall before submitting a pull request.

Maybe you want to explain how this reduces churn?

PatrickDropbox:
ah, much better (I only ran make tests before)

take a look at the test files.  we were importing the same packages multiple times with different name suffixes; these suffixes can change whenever the proto definitions are modified.  This results in a massive diffs whenever we recompile and check in proto files

protoc-gen-gogo/generator/helper.go
```diff
@@ -99,7 +99,10 @@ func (this *pluginImports) GenerateImports(file *FileDescriptor) {
 type Single interface {
 	Use() string
 	IsUsed() bool
-	Generate() string
+	Name() string
+	Location() string
+
+	Generate() string // DEPRECATED
```
awalterschulze: Please delete this method in the interface and in the implementation.
I don't like deprecated code lying around its not a necessity.

test/thetest.proto
```diff
@@ -583,10 +583,10 @@ message UnrecognizedWithEmbed {
 }
 
 //begin versioned: only for protoc 2.6.0 and up
-message SampleOneOf {
-  oneof test_oneof {
-    string name = 4;
-    NinOptNative sub_message = 9;
-  }
-}
+// // message SampleOneOf {
```
awalterschulze: I have made a fix for this that I haven't pushed yet.
My bad.
If possible could you please revert the change on this file.
Sorry.

awalterschulze:
This is a great patch.
Thank you very much.

PatrickDropbox:
rm deprecated function + reverted thetests.proto.

awalterschulze:
Thank you very much this is great :)

PR #45, opened 2015-04-10
Removed redundant and overzealous deduplicating of code for import statements

laurent-descrivan:
There is an issue regarding missing import statements when importing external .proto of a package, while also using a custom type of the very same package;

/main.proto

    package whatever

    import "custom/pkg/protofile.proto";

    message Object {
        optional bytes CustomField = 1 [(gogoproto.customtype) = "custom/pkg.CustomType"];
    }

/custom/pkg/protofile.proto

    package pkg;
    // whatever

/custom/pkg/customtype.go

    package pkg

    type CustomType struct { ... }

With this kind of structure, the generated main.pb.go will contain

    package whatever
    ....
    import pkg "custom/pkg"
    // import custom_pkg "custom/pkg" should also be present, but is not written
    ...
    type Object struct {
        CustomField *custom_pkg.CustomType `protobuf:....`
    }

In this generated main.pb.go, as the commented line is not present, the go compilation fails. I believe the code was to deduplicate the list of import statements, but fails to do so flawlessly, and is redundant since the commit 21e69275ee366ae7cebebe584f20ec360289337c. This pull request fixes that.

That would be great if the fix could be merged in the main repository, thanks. I am available for any question.


awalterschulze:
Could you please add a test in a new folder under the test folder?

laurent-descrivan:
I have added the test case, this fails for current master, and pass with the fix.

awalterschulze:
Thank you very much :)

laurent-descrivan:
Great, thanks ! I have another small fix on the way too. I'll make another PR when ready.

awalterschulze:
That is good news :)

Issue #36, opened 2014-12-22
Work In Progress: casttype (new type of customtype)


tv42:
main.go:
```go
package main

//go:generate protoc -I. -I$GOPATH/src -I$GOPATH/src/github.com/gogo/protobuf/protobuf --gogo_out=. repro.proto

type T uint32

func main() {
}
```

repro.proto:
```protobuf
package repro;

option go_package = "main";

import "github.com/gogo/protobuf/gogoproto/gogo.proto";

option (gogoproto.unmarshaler_all) = true;

message M {
  required uint32 X = 1 [(gogoproto.nullable) = false, (gogoproto.customtype) = "T"];
}
```

```console
$ go generate && go build
# _/home/tv/src/2014/protobuf-bug-type
./repro.pb.go:69: invalid operation: m.X |= uint32(b) & 127 << shift (mismatched types T and uint32)
```

tv42:
It seems just changing 

```go
				m.X |= (uint32(b) & 0x7F) << shift
```

to

```go
				m.X |= T((uint32(b) & 0x7F) << shift)
```

would be enough. I'm not sure if you want to keep this separate from `customtype`, which is always a length-prefixed []byte thing.

tv42:
Oh and MarshalBinary won't really solve this, because the wire format is different when the .proto says `uint32` vs `bytes`.

awalterschulze:
I agree MarshalBinary does not solve this, it might just make this use case less important.

awalterschulze:
We should also think about what happens when the type is another message and the field has a customtype.

For example:

message A {
  optional Uuid Id = 1 [(gogoproto.customtype="uuid")];
}

message Uuid {
  optional fixed64 First8Bytes = 1;
  optional fixed64 Second8Bytes = 2;
}

I have seen this exact use case in the wild.

This means that the go client would prefer just to work with the uuid type, but another client prefers the message type.

petermattis:
[...]

Issue #28, opened 2014-12-03
Vote For New Feature: Support BinaryMarshaler and TextMarshaler for custom types


awalterschulze:
_From [joa...@dynamic-design.se](https://code.google.com/u/116413275603785449127/) on September 29, 2014 21:40:54_

I have several times tried using custom types (mostly existing types from imported packages) with no luck. It always fails when I don't want to implement the proto.Marshaler interface, since I then have to wrap the type which bring (IMO) unnecessary complexity.

Since we can use a byte slice as a custom type, wouldn't it be perfectly logical to just use the existing go interface for marshalling to/from binary? ( http://golang.org/pkg/encoding/#BinaryMarshaler ) 

Also it should not be hard at all to implement, if you don't want this change maybe you could point me to the proper place for me to implement it locally?

_Original issue: http://code.google.com/p/gogoprotobuf/issues/detail?id=28_

tv42:
Personally I've never used Equal or Populate, but encounter the need for custom marshaling every now and then. For me, personally, it's not the end of the world if types using MarshalBinary won't work with Equal or Populate.

awalterschulze:
If customtype does not work with Equal and Populate is also does not work for any test generation.
I think this is a little scary, since your custom type is not tested anywhere else.

I think there are a couple of ideas about Equal, Marshaling and then checking, reflect.DeepEqual after finding that the type does not implement the equaler interface.

I hope testing/quick.Generator is used a lot in the standard library, since then this is a perfect replacement for Populate.

I just haven't had the time to check.

awalterschulze:
Ok quick.Generator does not look very popular in the standard library :(

grep -R "Generate(rand" .                         ./crypto/tls/handshake_messages_test.go:func (*clientHelloMsg) Generate(rand *rand.Rand, size int) reflect.Value {
./crypto/tls/handshake_messages_test.go:func (*serverHelloMsg) Generate(rand *rand.Rand, size int) reflect.Value {
./crypto/tls/handshake_messages_test.go:func (*certificateMsg) Generate(rand *rand.Rand, size int) reflect.Value {
./crypto/tls/handshake_messages_test.go:func (*certificateRequestMsg) Generate(rand *rand.Rand, size int) reflect.Value {
./crypto/tls/handshake_messages_test.go:func (*certificateVerifyMsg) Generate(rand *rand.Rand, size int) reflect.Value {
[...]

PR #46, opened 2015-04-14
Fixed generated unmarshalling of ints with custom types.

petermattis:
Previously the generated code wouldn't even compile. Not sure if this is the right approach. Note that the `Marshal` and `Unmarshal` methods are never called for int wire types when using the generated marshaler and unmarshaler.

### DIFF
--- a/plugin/unmarshal/unmarshal.go
+++ b/plugin/unmarshal/unmarshal.go
@@ -172,12 +172,13 @@ package unmarshal
 
 import (
 	"fmt"
+	"strconv"
+	"strings"
+
 	"github.com/gogo/protobuf/gogoproto"
 	"github.com/gogo/protobuf/proto"
 	descriptor "github.com/gogo/protobuf/protoc-gen-gogo/descriptor"
 	"github.com/gogo/protobuf/protoc-gen-gogo/generator"
-	"strconv"
-	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


awalterschulze:
CustomType currently only works for bytes.
This is a good idea that has been discussed in these issues.
https://github.com/gogo/protobuf/issues/28
https://github.com/gogo/protobuf/issues/36
I would really love to have this done for all scenarios, since I believe customtype is very useful for more than just bytes.
Maybe you would like to join a discussion so that we first design a good solution for this problem.

PS
Please do not import other projects, even in tests.
I am trying to keep the dependencies to a minimum.

awalterschulze:
I think my main problem was something with NewPopulated.
I think I can probably just let go and do something more practical.

petermattis:
Oops, removed the usage of the `cockroach/util` package. That was just some copy&paste of existing code.

I'll add a comment to #36 which is the exact issue this PR is addressing.

awalterschulze:
Cool :)

awalterschulze:
I hope the proposed casttype will fix this issue for you?
https://github.com/gogo/protobuf/issues/36

awalterschulze:
Please open again if you think this is still relevant.
You might be able to reuse some of this work if you would like to implement casttype.

PR #47, opened 2015-04-16
Fix go syntax generation for unmarshall when a non-struct custom type is used

laurent-descrivan:
There is an issue regarding unmarshalling of a custom type, is the type is not a struct, but an alias to anything else, like an int.

Test case and fix is included. Copyright of this patch is waived.

Thank you for merging this pull request, if possible.


awalterschulze:
This looks like a good fix.
Thank you very much.

laurent-descrivan:
Thanks for the merge!

awalterschulze:
Thanks for the fix :)

PR #48, opened 2015-04-29
Add checks in marshal/unmarshal for presence of required fields

### LINKED ISSUE #26: "required" spec ignored by Marshal, Unmarshal
_From [neil.conway@gmail.com](https://code.google.com/u/neil.conway@gmail.com/) on August 21, 2014 22:02:47_

The Marshal() and Unmarshal() methods generated by gogoprotobuf ignore the "required" specification. Although using "required" is generally discouraged, it can make sense in some circumstances -- and moreover, silently ignoring "required" (without prominent documentation) seems like not-great behavior.

Possible remedies:
1. Providing an option to have Marshal() and Unmarshal() obey the "required" directive.
2. Perhaps enabling #1 by default.
3. Prominently documenting this behavior, especially if #1 is not the default behavior.

_Original issue: http://code.google.com/p/gogoprotobuf/issues/detail?id=26_

jmtuley:
* Whitespace changes are the result of running `make all`
* Fixes #26
* Based on work previously started by @anton-povarov
* Includes tests in test/required

This will help Cloud Foundry's [Dropsonde](https://github.com/cloudfoundry/dropsonde) library to use the generated, non-reflective `Marshal`/`Unmarshal` functions safely.


test/issue8/proto.pb.go
```diff
@@ -64,6 +66,11 @@ func (m *Foo) Unmarshal(data []byte) error {
 		}
 		fieldNum := int32(wire >> 3)
 		wireType := int(wire & 0x7)
+
+		if fn := int(fieldNum - 1); fn <= 0 {
```
anton-povarov: generated code seems incorrect here.
fn <= 0 would lead to panic on next line.

jmtuley: @anton-povarov could you help us fix that, please? We copied the `Unmarshal` changes from [your suggestion](https://github.com/gogo/protobuf/compare/master...anton-povarov:unmarshaller_required_fields), as found in #26. I don't think I completely understand what's going on here. It looks like that line is being written by [this line](https://github.com/gogo/protobuf/compare/master...anton-povarov:unmarshaller_required_fields#diff-e0188cb7643b9790acd32abc395a5659R719).

Thank you!

anton-povarov: Just as a quick fix, i would generate something like (fn <= 0 && fn >=0), this is somewhat ugly, but should be enough to make it safe.

so the line becomes

p.P(`if fn := int(fieldNum-`, strconv.Itoa(rfMin), `); (fn >= 0 && fn <= `, strconv.Itoa(rfMax-rfMin), `) { `)

To explain what's going on: the code makes a bitmask of all fields, and sets bits to 1 for the fields that are present. Also tries to save on the size of a bitmask, by finding required fields with lowest id and highest id. rfMin/rfMax are those ids. Since array indexes start with 0, the code then adjusts incoming field ids into array offsets + bit offsets.

In hindsight this "optimization" seems unnecessary for 99.9% of uses, so it might be simpler to just remove it, and the code would become much simpler.
Another reasonable suggestion i can make - is move code that sets "this field is present" bits, into specific "case" statements, and generate the code doing it only for required fields and do all the math in advance, this should speed up the parsing code.

anton-povarov: Or we could do even better!
1. count all required fields + get their names
2. generate a struct inside each function with specific bits for each required field
3. generate code specific to the field to set the corresponding bit in that struct
4. generate code that check that all bits are set after parsing is done

This way - the bitmask would grow with the number of required fields, not max field id in the message.

Sorry that i'm just making suggestions here, and showing no code, got carried away :)

jmtuley: I think I get it. I'm going to try to state the full logic, to see if I'm right:

1. Generate a bitmask for keeping track of which required fields are present.
1. Unmarshal. At each field, fill out the mask if that field is required.
1. After unmarshaling, check the mask for any missing fields. If found, return an error.

Is that correct? If it is, may I ask what is the benefit of fully unmarshalling, instead of bailing the moment an error is found? The latter is the approach we took in our changes to `MarshalTo`, but maybe that was naïve.

anton-povarov: Yes, you understand correctly.

As for unmarshalling: i see no other way honestly. At the time we start parsing - we have no idea what's in the message. As fields come in - we get more info, but still can only answer the question "is this specific field present", but if it's required - that's what we want, and if it's not - we don't care.
And since fields are not always sent in order on the wire - we need to make sense of the whole message before checking.
An alternative approach might be to make 1st pass over the message - to get field types+offsets (handling sub-messages properly) then check and then 2nd pass to copy the data.
But it's much more complex and i would assume - slower for the common case when message is fully correct.

Anyway.
I've updated my implementation with the suggestion given in previous message (simpler code, should be more efficient when parsing as well, since we do more stuff at codegen time). Feel free to use it if you like.
https://github.com/anton-povarov/protobuf/commit/c65b7143b92c5e608cc236533e9dec374d238ec0

awalterschulze:
Sorry I have been a bit under the weather. It seems like this is on the right track.

@anton-povarov  that is super over optimised, but since one of gogoprotobuf's goals is being fast, I like, I like it a lot :)

@jmtuley could you please just call proto.Marshal and proto.Unmarshal in your tests?  These will call the generated code for you and malloc the appropriately sized byte slice.
Thank you for getting this rolling again :)

jmtuley:
@anton-povarov I cherry-picked your commit into my fork, so please take a quick look and verify that I did that correctly. :smile: Thank you for your help in getting that corrected and optimized!

@awalterschulze I've made the change you requested. I hope you're feeling better!

anton-povarov:
LGTM :)

awalterschulze:
The big picture LGTM.
Now I am just going to be a little pedantic about the tests and then we can merge :)

proto/encode.go
```diff
@@ -59,6 +59,10 @@ func (e *RequiredNotSetError) Error() string {
 	return fmt.Sprintf("proto: required field %q not set", e.field)
 }
 
+func NewRequiredNotSetError(field string) *RequiredNotSetError {
```
awalterschulze: Please move this to encode_gogo.go
I try to gogoprotobuf's extra code as seperate as possible from the original goprotobuf code.

awalterschulze:
The filemode should not need to change protoc-gen-gogo/main.go 100644 → 100755

plugin/unmarshal/unmarshal.go
```diff
@@ -827,6 +851,28 @@ func (p *unmarshal) Generate(file *generator.FileDescriptor) {
 		p.P(`}`)
 		p.Out()
 		p.P(`}`)
+
```
awalterschulze: Please remove all these newlines.

plugin/unmarshal/unmarshal.go
```diff
@@ -697,6 +711,7 @@ func (p *unmarshal) Generate(file *generator.FileDescriptor) {
 		if len(message.Field) > 0 {
 			p.P(`wireType := int(wire & 0x7)`)
 		}
+		p.P()
```
awalterschulze: remove this line please

plugin/unmarshal/unmarshal.go
```diff
@@ -740,6 +755,15 @@ func (p *unmarshal) Generate(file *generator.FileDescriptor) {
 				p.P(`}`)
 				p.field(field, fieldname)
 			}
+
+			if field.IsRequired() {
+				fieldBit, ok := rfMap[field.GetNumber()]
+				if !ok {
+					panic("field is required, but no bit registered")
+				}
+
```
awalterschulze: remove this line please

plugin/unmarshal/unmarshal.go
```diff
@@ -827,6 +851,28 @@ func (p *unmarshal) Generate(file *generator.FileDescriptor) {
 		p.P(`}`)
 		p.Out()
 		p.P(`}`)
+
+		p.P()
```
awalterschulze: remove this line please

test/example/example.pb.go
```diff
@@ -305,7 +308,8 @@ func (m *B) Unmarshal(data []byte) error {
 			if postIndex > l {
 				return io.ErrUnexpectedEOF
 			}
-			m.G = append(m.G, github_com_gogo_protobuf_test_custom.Uint128{})
```
awalterschulze: This is weird, did I not regenerate example before.
This seems like an old change?

test/required/requiredexample.proto
```diff
@@ -0,0 +1,14 @@
...
+
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.sizer_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+
+message RequiredExample {
+    required string theRequiredString = 1;
+    optional string theOptionalString = 2;
+    repeated string theRepeatedStrings = 3;
+}
```
awalterschulze: Please copy test.NidOptNative and test.NinOptNative and change optional to required.

Create a test that imports test.NidOptNative and calls test.NewPopulatedNidOptNative and Marshal the message.
Then take those bytes and Unmarshal them into required.NidOptNative.
This should never return an error.

Create another test that imports test.NinOptNative and calls test.NewPopulatedNinOptNative.
Now this might be a challenge.
I don't know how familiar you are with reflect, but I will write this test myself if you say that you don't have enough time.
Using reflect set a random field to Nil.
Then Marshal test.NinOptNative and Unmarshal the bytes into required.NinOptNative.
This should always return an error.

Finally Create a test using required.NinOptNative and required.NewPopulatedNinOptNative.
Again set a random field to Nil, easiest with reflect.
Then Marshal, this should always return an error.

test/required/requiredexamplepb_test.go
```diff
@@ -0,0 +1,69 @@
...
+	if err.Error() != `proto: required field "theRequiredString" not set` {
+		t.Fatalf(`err.Error() != "proto: required field "theRequiredString" not set"; was "%s" instead`, err.Error())
+	}
+}
+
+func TestUnmarshalSucceedsWhenRequiredIsNotPresent(t *testing.T) {
+	dataOut := RequiredExample{
+		TheRequiredString: proto.String("present"),
+	}
+	encodedMessage, _ := dataOut.Marshal()
```
awalterschulze: please check the error

test/required/requiredexamplepb_test.go
```diff
@@ -0,0 +1,69 @@
+package required
+
+import (
+	"github.com/gogo/protobuf/proto"
+	"testing"
+)
+
+func TestMarshalToErrorsWhenRequiredFieldIsNotPresent(t *testing.T) {
```
awalterschulze: less newlines would be nice :)

awalterschulze:
Thank you this really looks like it is going to work :)

test/example/example.pb.go
jmtuley: I'm not sure if it's an old change or not, but this is the code generated by `make all` on my machine (OS X 10.10.3, protoc 2.6.1 installed via homebrew). I can revert it.

test/required/requiredexamplepb_test.go
jmtuley: Sorry, I'm used to a coding style where we break up the code more. I apologize for not doing a better job conforming to the style in the library.

awalterschulze: Its not really a big problem at all :)

test/example/example.pb.go
awalterschulze: I regenerated the code on master now.
So if you pull in master again, this diff should go away.
My bad.

test/required/requiredexample.proto
jmtuley: I'm unable to do exactly as you ask.

When I add a message
```protobuf
message NidOptNative {
	required double Field1 = 1 [(gogoproto.nullable) = false];
}
```
to requiredexample.proto, I get the following compile error:
```
 make
(protoc --proto_path=../../../../../:../../protobuf/:. --gogo_out=. requiredexample.proto)
2015/05/05 08:18:56 protoc-gen-gogo: error:bad Go source code was generated: 1070:57: expected operand, found ']' (and 2 more errors)
--gogo_out: protoc-gen-gogo: Plugin failed with status code 1.
make: *** [regenerate] Error 1
```

Should I just remove the `[(gogoproto.nullable) = false]` tags (which would make `required.NidOptNative` identical to `required.NinOptNative`), or do you want to figure out why `required` fields clash with setting `nullable` to `false`?

awalterschulze: I just tried that.
This seems to only be an error on this branch.
So I suspect it has to do with the newly generated code.
I suspect we have found a bug :)

jmtuley: Since the generated Go source is discarded, do you have any recommendation for how to debug this bug?

Thanks!

awalterschulze: I think I might have enabled this on the proto3 branch.

```
	rem := g.Buffer
	g.Buffer = new(bytes.Buffer)
	g.generateHeader()
	g.generateImports()
	g.Write(rem.Bytes())

	// Reformat generated code.
	fset := token.NewFileSet()
	ast, err := parser.ParseFile(fset, "", g, parser.ParseComments)
	if err != nil {
		g.Fail("bad Go source code was generated:", err.Error())
		return
	}
```

Simply print out g.Buffer.Bytes() if the error occurs.

awalterschulze:
How is it going?

geapi:
We are about to finish the last two tests you requested.

awalterschulze:
cool I am very excited :)

geapi:
@awalterschulze 

Thanks for your help and direction. With the last commit, we've added the requested tests. Is there anything else you'd like from us, or is this ready to merge?

awalterschulze:
The tests look great :)

I am just not sure about the new map, instead of the previous probably more efficient bit implementation.
I don't think it needs to be that efficient, but a map is very inefficient.
Why did you change it?

anton-povarov:
Same question here, adding more allocations and hash lookups to a function that should be fast is counter intuitive imo.

jmtuley:
We changed it due to a bug in the other implementation that we couldn't
figure out.

As we wrote the tests, we found that after field 6, all the masks were 0,
which obviously want going to work. Based on Walter's comment that a bit
mask was probably over optimized, we chose to solve the issue with
straightforward code, i.e. the map implementation, that we could understand
and didn't get in the way if getting the tests green.

There's also not that much extra allocation, since we allocate the whole
hash at once. Using a map allowed us to range at the final check, which
avoids the bound checking that happens in array lookups.

I don't think we are likely to correctly implement the mask logic, but
won't be offended if someone does manage it and adds the commit to our pull
request. I'm sorry that we were unable to correct the initial
implementation.

awalterschulze:
Ok no problem.  I'm sure I (if Anton doesn't have time) can optimise this.

I was wondering if you could add one more test please.
I want to test nested messages.

So just add a new message which has a repeated NinOptNative.
And have one of those messages not the last or the first, have a nil field which is required.

Marshal and/or Unmarshal should give an error, which ever you have time for.

If you add the final test, I (if Anton doesn't have time) wil optimise the map.

anton-povarov:
Please do point at the test that is failing, i'll fix.

jmtuley:
I'm afraid I'm not in a position to do much right now; I just had wrist
surgery and so can't type effectively. (This is being painfully tapped out
on a phone.)

I'll do more on Thursday when the bandages come off.

anton-povarov:
Righto, this one turned out to be pretty nasty, but easier than figuring github out for a newbie like me :)
The fix: https://gist.github.com/anton-povarov/c485ded98b51cb6605a0

The issue was with operator precedence, << has higher precedence than %.
so for fields 6 and above it would first shift left (making the value a multiple of 64 and always larger) and then take remainder, making it 0.

This diff also changes number formatting for bits set/check from decimal to hex, i find it easier to read and the code is less cluttered with parens.

NB @awalterschulze : all_test.go:1283 fails go vet check on master :)

anton-povarov:
As to performance comparison, figured i'd just benchmark it the regular way.
Can't say i've spent too much time on it, just keps on copy-pasting code from `test/required/requiredexample.pb.go` till it stopped complaining.

```
antoxa@antoxa-suse:~/_Dev/go/src/antoxa> go test -bench . -benchtime 10s ./1_test.go
testing: warning: no tests to run
PASS
BenchmarkMap     2000000              6644 ns/op
BenchmarkBitmask        20000000               989 ns/op
ok      command-line-arguments  41.237s
```

full code: https://gist.github.com/anton-povarov/6a3733d77da9ac837ba9

jmtuley:
Well, that's a little faster. :-) Good work!

geapi:
We applied Anton's patch and added a test for the nested NinOptNative, please let us know if there is anything else we can do.

awalterschulze:
Ok this looks perfect! :)

awalterschulze:
@anton-povarov my go vet works, which go version are you running?

awalterschulze:
I am running go version go1.4 darwin/amd64

anton-povarov:
1.4.2
```
antoxa@antoxa-suse:~/_Dev/go/src/github.com/gogo/protobuf> go vet ./...
proto/all_test.go:1283: result of fmt.Sprintf call not used
exit status 1
antoxa@antoxa-suse:~/_Dev/go/src/github.com/gogo/protobuf> go version
go version go1.4.2 linux/amd64

awalterschulze:
That should also be an issue in golang/protobuf.
You could report it there and I'll merge it when they fix it or even if they refuse to fix it.

awalterschulze:
Awesome Thank you for everyone's work.
I really appreciate it.
This was great team work :+1:

anton-povarov:
Glad to see this merged, thanks guys :)

PR #50, opened 2015-05-07
Simple proof of concept for allowing aliased but non-marshaled types

laurent-descrivan:
This is a simple unfinished proof of concept for illustration in issue #36


awalterschulze:
I will work on a solution for this asap.
I know that casted types is a usecase that is quite popular.
So I will definitely be looking to solve it.
This would only be the start to the solution, there is still a lot to change in proto/decode and proto/encode etc.
I should have already designed something, I'm sorry, but please be patient.
Once I have designed something I will be needing help.
So keep on watching issue 36

Issue #64, opened 2015-06-08
casttype doesn't handle nullable []byte correctly


tamird:
In proto2 syntax:
```proto
message RangeTreeNode {
  optional bytes left_key = 4 [(gogoproto.casttype) = "Key"];
}
```
In the generated code, LeftKey should be a pointer (because this thing is nullable) but it is not. Even the more explicit:
```proto
message RangeTreeNode {
  optional bytes left_key = 4 [(gogoproto.nullable) = true, (gogoproto.casttype) = "Key"];
}
```
yields a value type rather than a pointer.

awalterschulze:
This you can only do with customtype.
casttype assumes that the generated code without casttype is castable to the casttype.
bytes does not generate *[]byte, but generates []byte, so it is not castable to *Key

PR #65, opened 2015-06-08
Address #64, sort of

tamird:
Tests don't pass, and I'm not sure this is the right way to make this change, but it does generate code that compiles.


awalterschulze:
Yes thats because the generated code is just one facet.
The marshaling and unmarshaling should also work when no code generation is used.
And that is basically the big reason I don't want to support *[]byte.
The reflect and unsafe marshaling and unmarshaling of *[]byte instead of []byte will be hard to implement.
Next it is a very special case for bytes, none of the other field types need this.
And you can basically already implement this with customtype if you really need a pointer.

PR #67, opened 2015-06-12
Support multiple packages per run

tamird:
This is a modified version of https://github.com/golang/protobuf/pull/40. Upstream seems to be reluctant to support non-Google use-cases :(


awalterschulze:
Seems to me that his concern was more about Go packages than Google?

tamird:
I think the TL;DR was that they don't need it, so they don't want to support it. In any case, how does this change look to you?

tamird:
One thing I couldn't quite work out is that this changes seems to be overly-eager with renaming imports:
```diff
-import proto1 "github.com/gogo/protobuf/proto"
+import proto11 "github.com/gogo/protobuf/proto"
```
But I couldn't figure out the renaming logic (which is different from upstream).

tamird:
Figured it out! PTAL.

awalterschulze:
I think what I need is a very clean example of how you would use this.
I have looked at the cockroachdb issue, but I can't seem to wrap my head around it.

Maybe you could make a folder in the test folder with some subfolders and some protos with a Makefile in the test/yourfolder that protocs all the files with one call.
This way we have a test to make sure it keeps on working.

tamird:
You got it - added the same changes I made in https://github.com/golang/protobuf/pull/40 to exercise this new code.

awalterschulze:
I would prefer a clean example if that would be possible?  
  - I like to minimize the changes in overlapping files between golang/protobuf and gogo/protobuf
  - And then it should be easier for me to play around with, since there will be no other factors.

tamird:
Done, made a freestanding test case.

awalterschulze:
Could you please move the multipkg folder from the testdata folder to the test folder?
Could you also add a Makefile in the root of the multipkg folder to show how you use your modifications to generate all the files?

tamird:
Done!

protoc-gen-gogo/generator/generator.go
```diff
@@ -549,6 +548,19 @@ var uniquePackageName = make(map[*descriptor.FileDescriptorProto]string)
 // value is the name that appears in the generated code.
 var pkgNamesInUse = make(map[string][]*FileDescriptor)
 
+func pkgNameInUse(pkg string, f *FileDescriptor) bool {
```
awalterschulze: Seems this is only a cleanup of the old code.
Could you revert the use of this function so that the diff with golang/protobuf is as small as possible.

test/multipkg/bar/bar.proto
```diff
@@ -0,0 +1,41 @@
+// Extensions for Protocol Buffers to create more go like structures.
```
awalterschulze: Rather use this copyright, since its gogoprotobuf original

// Copyright (c) 2015, Vastech SA (PTY) LTD. All rights reserved.
// http://github.com/gogo/protobuf/gogoproto
//
// Redistribution and use in source and binary forms, with or without
// modification, are permitted provided that the following conditions are
// met:
//
//     * Redistributions of source code must retain the above copyright
// notice, this list of conditions and the following disclaimer.
//     * Redistributions in binary form must reproduce the above
// copyright notice, this list of conditions and the following disclaimer
// in the documentation and/or other materials provided with the
// distribution.
//
// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.

test/multipkg/baz/baz.proto
```diff
@@ -0,0 +1,41 @@
+// Extensions for Protocol Buffers to create more go like structures.
+//
+// Copyright (c) 2013, Vastech SA (PTY) LTD. All rights reserved.
```
awalterschulze: ditto copyright

test/multipkg/foo/foo.proto
```diff
@@ -0,0 +1,41 @@
+// Extensions for Protocol Buffers to create more go like structures.
+//
+// Copyright (c) 2013, Vastech SA (PTY) LTD. All rights reserved.
+// http://github.com/gogo/protobuf/gogoproto
+//
+// Go support for Protocol Buffers - Google's data interchange format
```
awalterschulze: Ditto copyright

protoc-gen-gogo/generator/generator.go
tamird: it's not cleanup - i had to add some logic to see here. This is actually a reduction in the diff with golang/protobuf - take a look at the implementation of `RegisterUniquePackageName` there.

tamird:
Updated the copyright

protoc-gen-gogo/generator/generator.go
awalterschulze: My bad

awalterschulze: I really suck at reading code.

protoc-gen-gogo/generator/generator.go
```diff
@@ -635,45 +635,49 @@ func (g *Generator) defaultGoPackage() string {
 // The package name must agree across all files being generated.
 // It also defines unique package names for all imported files.
 func (g *Generator) SetPackageNames() {
-	// Register the name for this package.  It will be the first name
-	// registered so is guaranteed to be unmodified.
-	pkg, explicit := g.genFiles[0].goPackageName()
```
awalterschulze: Why is the explicit goPackageName not needed anymore?

tamird: because we have a mapping from each file to its go package name.

awalterschulze: So this function or something like it is being called somewhere else
```
// goPackageName returns the Go package name to use in the
// generated Go file.  The result explicit reports whether the name
// came from an option go_package statement.  If explicit is false,
// the name was derived from the protocol buffer's package statement
// or the input file name.
func (d *FileDescriptor) goPackageName() (name string, explicit bool) {
	// Does the file have a "go_package" option?
	if opts := d.Options; opts != nil {
		if pkg := opts.GetGoPackage(); pkg != "" {
			return pkg, true
		}
	}

	// Does the file have a package clause?
	if pkg := d.GetPackage(); pkg != "" {
		return pkg, false
	}
	// Use the file base name.
	return baseName(d.GetName()), false
}
```

Aha I see that is what opts.GetGoPackage() is doing.

awalterschulze:
The github app is giving weird error:
```
From https://github.com/gogo/protobuf
 ! [rejected]        refs/pull/67/head -> pr/67  (non-fast-forward)
 (1)
```
When I clone your protobuf-1 repo I also can't find the test/multipkg folder.
I wanted to change the test a bit and play around.
Any ideas?
Maybe I am just tired and should try again tomorrow.

tamird:
@awalterschulze in my repo it's not on master, it's on the `multiple-packages-per-run` branch.

awalterschulze:
I made some changes to test/multipkg/Makefile, but could not get it to work.
The current import "baz/baz.proto" confuses go.
While this import "github.com/gogo/protobuf/test/multipkg/baz/baz.proto" confuses protoc.

```
test:
	make testprotoc
	make regenerate
	go test -v ./...

regenerate:
	(protoc --proto_path=../../../../../:../../protobuf/:. --gogo_out=. $(shell find . -name *.proto -type f))

testprotoc:
	(protoc --proto_path=../../../../../:../../protobuf/:. --descriptor_set_out=/dev/null $(shell find . -name *.proto -type f))	
```

test/multipkg/baz.proto
```
syntax = "proto2";

option (gogoproto.goproto_stringer_all) = false;
option (gogoproto.goproto_enum_prefix_all) = false;
option (gogoproto.goproto_getters_all) = false;

option (gogoproto.equal_all) = true;
option (gogoproto.verbose_equal_all) = true;
option (gogoproto.stringer_all) = true;
option (gogoproto.gostring_all) = true;
option (gogoproto.face_all) = true;
option (gogoproto.description_all) = true;

option (gogoproto.benchgen_all) = true;
option (gogoproto.unmarshaler_all) = false;
option (gogoproto.marshaler_all) = false;

option (gogoproto.goproto_enum_stringer_all) = false;
option (gogoproto.enum_stringer_all) = true;

option (gogoproto.unsafe_marshaler_all) = false;
option (gogoproto.unsafe_unmarshaler_all) = false;

package baz;

message BazMessage {
	optional int64 BazField = 1;
}
```

test/multipkg/bar.proto
```
syntax = "proto2";

import "baz/baz.proto";

option (gogoproto.goproto_stringer_all) = false;
option (gogoproto.goproto_enum_prefix_all) = false;
option (gogoproto.goproto_getters_all) = false;

option (gogoproto.equal_all) = true;
option (gogoproto.verbose_equal_all) = true;
option (gogoproto.stringer_all) = true;
option (gogoproto.gostring_all) = true;
option (gogoproto.face_all) = true;
option (gogoproto.description_all) = true;

option (gogoproto.benchgen_all) = true;
option (gogoproto.unmarshaler_all) = false;
option (gogoproto.marshaler_all) = false;

option (gogoproto.goproto_enum_stringer_all) = false;
option (gogoproto.enum_stringer_all) = true;

option (gogoproto.unsafe_marshaler_all) = false;
option (gogoproto.unsafe_unmarshaler_all) = false;

package bar;

message BarMessage {
	optional baz.BazMessage BarMsg = 1;
}
```

awalterschulze:
I added all the extensions basically just so I could have tests to run.  So most of them are unnecessary.
I think this shows some of the mismatch between protoc and go.
Maybe you have a solution for this situation?

tamird:
Yeah, nice catch. Basically, protoc-gen-go cannot know about the "real" final name of the go package, so it can't generate the correct import path. Ugh. We could possibly inspect the relative path from GOPATH to the current directory, but that is pretty gross.

Got any ideas?

awalterschulze:
Unfortunately not.
I agree using the $GOPATH is very gross.

awalterschulze:
I think it should be pretty easy to write a script that knows about the GOPATH and how to use protoc and basically scan all folders to find proto files and run protoc accordingly.
I think this is something that is built on top of protoc and some go support protobuf.

tamird:
Yeah, that's what we ended up doing in https://github.com/cockroachdb/cockroach/commit/f1b7cd32c9961b839c795bda6a33f6f5b6bf8758

Do you want to include that in gogoproto in some way?

awalterschulze:
It depends on what that some way is.
This would need to be a separate go binary, that calls protoc for you.
It should be possible to also include go_out or gogo_out or gofast_out, etc as a parameter.

awalterschulze:
Please open a new issue or pull request when you have a new proposal.

PR #83, opened 2015-07-17
grpc code generation workaround

andres-erbsen:
Sometimes code generation with `--gogofast_out=plugins=grpc:.` fails because the statements the grpc plugin adds to avoid "unused import" errors end up being in the middle of the import list, causing the build to fail regardless of whether the imports are used or not. This pull request fixes grpc code generation for the case where the imports are used by removing the `_ = pkg.Func` statements from grpc code generation, thus totally breaking the case where they are not used.

The issue was observed with go1.5 from Wednsday (2e4b659) and the protobufs in https://github.com/yahoo/coname/tree/master/proto.

### DIFF
--- a/plugin/grpc/grpc.go
+++ b/plugin/grpc/grpc.go
@@ -113,10 +113,6 @@ func (g *grpc) GenerateImports(file *generator.FileDescriptor) {
 	g.P(grpcPkg, " ", strconv.Quote(path.Join(g.gen.ImportPrefix, grpcPkgPath)))
 	g.P(")")
 	g.P()
-	g.P("// Reference imports to suppress errors if they are not otherwise used.")
-	g.P("var _ ", contextPkg, ".Context")
-	g.P("var _ ", grpcPkg, ".ClientConn")
-	g.P()
 }
 
 // reservedClientName records whether a client name is reserved on the client side.
--- a/test/issue83/Makefile
+++ b/test/issue83/Makefile
@@ -1,0 +1,30 @@
+# Extensions for Protocol Buffers to create more go like structures.
+#
+# Copyright (c) 2013, Vastech SA (PTY) LTD. All rights reserved.
+# http://github.com/gogo/protobuf
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are
+# met:
+#
+#     * Redistributions of source code must retain the above copyright
+# notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above
+# copyright notice, this list of conditions and the following disclaimer
+# in the documentation and/or other materials provided with the
+# distribution.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+# A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+# OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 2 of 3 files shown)


awalterschulze:
Maybe you could provide a smallest example where the error occurs, that way we can have a regression test to make sure this does not happen again in future.
Add another folder to the test folder containing the proto with a makefile.  See the other test folders for examples.
Or if this is too much trouble I would really just like to have a smallest example proto please.
Then I can write the test, etc.

Also are you ok with giving over your copyright on this pull request?

Would you mind if I add coname to the list of gogoprotobuf users?

andres-erbsen:
The copyright to this code is owned to Yahoo and, I talked to the relevant people, the company not willing to give it away. However, you are free to use the code under the license in the very beginning of the file. If that does not work for you, please consider this pull request just a bug report (I did update my branch with a minimal example and a Makefile). I hope this is not too much trouble.

awalterschulze:
Ok I am then taking this as a bug report.

When I run the makefile on your proto I get the following

```
...
import (
	context "golang.org/x/net/context"
	grpc "google.golang.org/grpc"
)

// Reference imports to suppress errors if they are not otherwise used.
var _ context.Context
var _ grpc.ClientConn
...
```

awalterschulze:
I don't see the problem here.

awalterschulze:
Maybe I am not understanding your problem?
But everything is compiling.

awalterschulze:
Please reopen an issue if this is still a problem.

andres-erbsen:
Did you test using go1.5? Using the version referenced in my first post, I get the following output:

~~~~~~~~~~~~~~~~
uhu@ashryn ~/go/src/github.com/yahoo/coname/proto (git)-[master] % grep -C1 'var _' /home/uhu/go/src/github.com/gogo/protobuf/plugin/grpc/grpc.go                         :(
	g.P("// Reference imports to suppress errors if they are not otherwise used.")
	g.P("var _ ", contextPkg, ".Context")
	g.P("var _ ", grpcPkg, ".ClientConn")
	g.P()
uhu@ashryn ~/go/src/github.com/yahoo/coname/proto (git)-[master] % go version                                                                    
go version devel +2e4b659 Wed Jul 15 06:04:51 2015 +0000 linux/amd64
uhu@ashryn ~/go/src/github.com/yahoo/coname/proto (git)-[master] % protoc --gogo_out=plugins=grpc:. -I . -I "$GOPATH/src/github.com/gogo/protobuf" *.proto

... (generated code dump)

 48:110: expected declaration, found 'import'
--coname_out: protoc-gen-coname: Plugin failed with status code 1.
~~~~~~~~~~~~~~~~

andres-erbsen:
The generated code has some import statements, some "var _ = pkg.Func" statements, and then some more import statements. This does not seem to be valid Go by https://golang.org/ref/spec#Source_file_organization.

awalterschulze:
No I am waiting for go1.5 to come out.

What is this?
--coname_out: protoc-gen-coname: Plugin failed with status code 1.
Aha I am just going to put this here as a reference
https://github.com/yahoo/coname/blob/5c94d9ddbae714e7d08d0673eecb2311a2e21924/proto/protoc-gen-coname/main.go
I see you also like your testgen and gostring, yeah :)

I'll try to reproduce this again, now that I know about the import statement ordering problem.

awalterschulze:
I get the following which seems correct.

```
package issue83

import proto "github.com/gogo/protobuf/proto"

import io "io"
import fmt "fmt"

import (
	context "golang.org/x/net/context"
	grpc "google.golang.org/grpc"
)

// Reference imports to suppress errors if they are not otherwise used.
var _ context.Context
var _ grpc.ClientConn

// Reference imports to suppress errors if they are not otherwise used.
var _ = proto.Marshal

```

But I see the problem.
You should be able to produce the same problem with golang/protobuf, since this is just their forked code.
Maybe you would like to report and issue there?
Then I could just pull in the fix, instead of making my own fix and then one day reverting my fix and merging in their fix.

awalterschulze:
This should mostly fix your problem
https://github.com/gogo/protobuf/commit/3bbffe5c32dbc6d7791590439d6151b130a8f945

awalterschulze:
Sorry my bad, here is the commit
https://github.com/gogo/protobuf/commit/cabd153b69f71bab8b89fd667a2d9bb28c92ceb4

PR #89, opened 2015-08-08
Add vet --shadow

tamird:
Just another thing I fixed a few times; wanted to make sure it's in CI :)


protoc-gen-gogo/generator/helper.go
```diff
@@ -247,11 +248,11 @@ func (g *Generator) generatePlugin(file *FileDescriptor, p Plugin) {
 	}
 }
 
-func GetCustomType(field *descriptor.FieldDescriptorProto) (packageName string, typ string, err error) {
+func GetCustomType(field *descriptor.FieldDescriptorProto) (_packageName string, _typ string, _err error) {
```
awalterschulze: please no _varname

fieldpath/fieldpath-generated.go
```diff
@@ -266,10 +266,11 @@ func (this *Float64SinglePath) UnmarshalFirst(buf []byte) (*float64, error) {
 				endOfs[position] = offset + int(length)
 			}
 		} else {
-			offset, err = skip(buf, offset, wireType)
+			newOffset, err := skip(buf, offset, wireType)
```
awalterschulze: This is generated, is the generator also changed?

Makefile
```diff
@@ -94,7 +94,8 @@ tests:
 	go test ./...
 
 vet:
-	go vet ./...
+	go tool vet --all .
+	! go tool vet --shadow . 2>&1 | grep -vE 'declaration of err shadows declaration at (proto/text.go:535|proto/text.go:570|protoc-gen-gogo/main.go:84)'
```
awalterschulze: What is happening over here?

awalterschulze:
Build is not passing.

protoc-gen-gogo/generator/helper.go
tamird: would you prefer I remove these entirely? they aren't used in `return`s

Makefile
tamird: it's a whitelist for stuff I didn't feel like fixing - consider it a todo list

fieldpath/fieldpath-generated.go
tamird: fixed now

awalterschulze: cool

Makefile
awalterschulze: I don't really like TODOs in a mature open source product.  Seems like its only 3 errors, why not just finish what you started.

tamird: The problem is that those are false positives; see https://github.com/golang/go/issues/11843

awalterschulze: Ok well maybe we shouldn't be using a tool for automated checking if it is producing false positives.

tamird: That's throwing away the baby with the bathwater. It mostly works, and to work around these false positives is pretty trivial. What's the harm?

awalterschulze: Whats the harm is just waiting until it is fixed?
The harm in committing this now is having the next committers possibly struggle around false positives and having to explain to them how to add their exceptions to the list.

fieldpath/fieldpath-gen/template.go
```diff
@@ -1121,10 +1121,11 @@ func (this *{{.CName}}SinglePath) UnmarshalFirst(buf []byte) ({{.ReturnType}}, e
 				endOfs[position] = offset + int(length)
 			}
 		} else {
-			offset, err = skip(buf, offset, wireType)
+			newOffset, err := skip(buf, offset, wireType)
```
awalterschulze: Isn't this also weird.
I am allowed to the offset, below, but not at the actual place I want to set it?

Makefile
tamird: > Whats the harm is just waiting until it is fixed?

It's probably going to be some time before that happens.

> The harm in committing this now is having the next committers possibly struggle around false positives and having to explain to them how to add their exceptions to the list.

Yeah, that's a risk. On the other hand, it can prevent bugs like this one: https://github.com/gogo/protobuf/pull/89/files#diff-9878bbf47ee5b9142c5e6fe30f61cbabR255

fieldpath/fieldpath-gen/template.go
tamird: this is a workaround for a false positive (see the commit message). the problem is that assigning to a previously defined `err` is ambiguous because it is shadowed above.

protoc-gen-gogo/generator/helper.go
```diff
@@ -253,7 +254,8 @@ func GetCustomType(field *descriptor.FieldDescriptorProto) (packageName string,
 
 func getCustomType(field *descriptor.FieldDescriptorProto) (packageName string, typ string, err error) {
 	if field.Options != nil {
-		v, err := proto.GetExtension(field.Options, gogoproto.E_Customtype)
+		var v interface{}
```
tamird: note that before this fix, this function would never return a non-nil error.

Makefile
awalterschulze: Nice :)
Why not make a separate make vetshadow that is not part of the automated make all.
That why we can manually choose to run the tool to find bugs, but not pressure people into working around a tool that is still experimental.

tamird: heh, if it's not in the automated build it may as well be left out entirely. it will not provide any help in preventing these bugs (unless you're planning on running `make vetshadow` yourself on every PR?)

fieldpath/fieldpath-gen/template.go
awalterschulze: I would like not to have any workarounds for false positives.
If this is the case, then this is exactly what I would like to avoid.

protoc-gen-gogo/generator/helper.go
awalterschulze: Except if field.Options == nil

Makefile
awalterschulze: Basically the idea would be to run it manually some of the time and decide if the shadows are real bugs.  The next step would be to ONLY fix those that are real bugs.  Since the tool can only provide hints, at this point, as to whether variables are shadowed or not, it can not be incorporated as an automated tool.

fieldpath/fieldpath-gen/template.go
tamird: actually, this one is easier to fix than i thought. Fixed in a different way.

Makefile
tamird: ok, i put it in a separate target.

everything in `test/` is passing, so that is still being checked in CI.

Makefile
```diff
@@ -94,7 +94,12 @@ tests:
 	go test ./...
 
 vet:
-	go vet ./...
+	go tool vet --all .
+	go tool vet --shadow ./test
```
awalterschulze: Please remove this line as well.  Even though it is working in the test folder now.  I don't want future commits struggling with this tool until the issue is sorted.

Makefile
```diff
@@ -94,7 +94,12 @@ tests:
 	go test ./...
 
 vet:
-	go vet ./...
+	go tool vet --all .
```
awalterschulze: Does vet run recursively, by default.
Shouldn't it be ... instead of .

tamird: https://github.com/golang/go/issues/11727

awalterschulze: Ok why not just go vet ... then?
Does go tool vet --all check more stuff?

tamird: for symmetry with the other invocations; no other reason

awalterschulze: Ok well nothing except the manual vet is using the go tool invocation.  Could please change it back, sorry.

tamird: done

Makefile
```diff
@@ -95,6 +95,11 @@ tests:
 
 vet:
 	go vet ./...
+	go tool vet --shadow ./test
```
awalterschulze: Just this one line still needs to be removed.

tamird: I insist that leaving this out is asking for more work. Feel free to commit the other changes here if you want, but I won't remove this.

awalterschulze: I am sorry about all the bickering, but I feel that when tools are in an experimental phase they cause pain when they are built into a process.
The pain is only worth it, if this is contributing back to the experiment, which I don't think it is.
The tool is valuable and finds bugs, as you have shown, but it requires human intervention, at this point, that is why it will be great to run it manually.
Please take your time and reconsider.

tamird: removed

awalterschulze:
Thank you very much.
Sorry for the tediousness of this merge.

PR #95, opened 2015-08-19
Sort plugins before codegen

### LINKED ISSUE #92: generated code changes when plugins are compiled with go1.5
Looks like this is because plugins are registered using `init()`, and the load order has changed in 1.5, but I'm not sure.

tamird:
Fixes #92.

This diff is **big**. Sorry about that.


BramGruneir:
LGTM

awalterschulze:
Could you explain why the sort is needed.
I am a bit tired, but I don't see any reason that the plugin list would not be deterministic.
Yes I still have not installed go1.5, but just looking at the code, I can't see what is going wrong.

awalterschulze:
Basically I think the change is fine, but I just want to understand exactly why it is needed first.

tamird:
The order of init() calls seems to have changed in 1.5.

tamird:
To clarify, it is still deterministic in 1.5, but the order differs from
1.4's. This ensures that both versions will produce the same code.

protoc-gen-gogo/generator/generator.go
```diff
@@ -77,7 +78,21 @@ type Plugin interface {
...
+
+func (ps pluginSlice) Less(i, j int) bool {
+	return ps[i].Name() < ps[j].Name()
+}
+
+func (ps pluginSlice) Swap(i, j int) {
+	ps[i], ps[j] = ps[j], ps[i]
+}
+
+var plugins pluginSlice
```
awalterschulze: This can change back to []Plugin

awalterschulze:
Everything else looks good and I understand now.   For some reason I thought the range was happening over pluginList and not plugins, as I said I am tired.

awalterschulze:
Thanks again for this fix :)

tamird:
My pleasure!

PR #98, opened 2015-08-27
Fix bug where package name with multiple segments broke codegen

tamird:
@awalterschulze this is the bug we saw over in https://github.com/cockroachdb/cockroach/pull/2276. See the change in `test/mapsproto2/combos/both/mapsproto2.proto`: adding the period breaks codegen!


tamird:
I believe I have found the source of this problem and have a fix.

PR #104, opened 2015-09-13
Added a method 'Comments' to generator.

ceram1:
Requirement: I need to print comment as quoted string.

### DIFF
--- a/protoc-gen-gogo/generator/generator.go
+++ b/protoc-gen-gogo/generator/generator.go
@@ -1183,12 +1183,26 @@ func (g *Generator) generateHeader() {
 // The path is a comma-separated list of integers.
 // See descriptor.proto for its format.
 func (g *Generator) PrintComments(path string) {
-	if loc, ok := g.file.comments[path]; ok {
-		text := strings.TrimSuffix(loc.GetLeadingComments(), "\n")
-		for _, line := range strings.Split(text, "\n") {
-			g.P("// ", strings.TrimPrefix(line, " "))
-		}
+	text := g.Comments(path)
+	if text == "" {
+		return
 	}
+
+	for _, line := range strings.Split(text, "\n") {
+		g.P("// ", strings.TrimPrefix(line, " "))
+	}
+}
+
+// Comments returns any comments from the source .proto file and empty string if comments not found.
+// The path is a comma-separated list of intergers.
+// See descriptor.proto for its format.
+func (g *Generator) Comments(path string) string {
+	loc, ok := g.file.comments[path]
+	if !ok {
+		return ""
+	}
+	text := strings.TrimSuffix(loc.GetLeadingComments(), "\n")
+	return text
 }
 
 func (g *Generator) fileByName(filename string) *FileDescriptor {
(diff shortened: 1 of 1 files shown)


protoc-gen-gogo/generator/generator.go
```diff
@@ -1183,12 +1183,26 @@ func (g *Generator) generateHeader() {
...
+
+	for _, line := range strings.Split(text, "\n") {
+		g.P("// ", strings.TrimPrefix(line, " "))
+	}
+}
+
+// Comments returns any comments from the source .proto file and empty string if comments not found.
+// The path is a comma-separated list of intergers.
+// See descriptor.proto for its format.
+func (g *Generator) Comments(path string) string {
```
awalterschulze: What about maybe return the ok and the string, just in the case that someone has an empty comment, which is weird, but not impossible I guess?

awalterschulze:
Please resolve the conflicts.

awalterschulze:
What are you planning to use this for?
I am quite excited to use this for creating help messages for letmegrpc.

protoc-gen-gogo/generator/generator.go
```diff
@@ -1183,12 +1183,26 @@ func (g *Generator) generateHeader() {
 // The path is a comma-separated list of integers.
 // See descriptor.proto for its format.
 func (g *Generator) PrintComments(path string) {
-	if loc, ok := g.file.comments[path]; ok {
-		text := strings.TrimSuffix(loc.GetLeadingComments(), "\n")
-		for _, line := range strings.Split(text, "\n") {
-			g.P("// ", strings.TrimPrefix(line, " "))
-		}
+	text := g.Comments(path)
```
awalterschulze: I think its import that the ok is accessible here.
I don't mind a little code duplication.
Copy the Comments method as is to helper.go file, because I think it actually has a nice API, but the PrintComments method requires the ok.
Revert the changes to PrintComments and leave it as it was.
I think this gives us the best of both worlds, what do you think?

awalterschulze:
Are you ok with giving over your copyright on this work?

protoc-gen-gogo/generator/generator.go
ceram1: I agree with it. I reverted it at https://github.com/ceram1/protobuf/commit/54e5631f0de9fbab0fdf83f008494a5be49dbb0e

ceram1:
I'll use protobuf as api declaration spec, as commenting for generated documentation in golang becomes 
```golang
// No way to comment here!
type GetResourceRequest struct {
    ID int64 `comment:"ID of the resource to get."`
}
```
but as I use protobuf, it becomes
```protobuf
// Request to get resource
message GetResourceRequest {
    // ID of the resource to get.
    int64 id = 1;
}
```
and I can emit other language stuffs, or even static documentations.

+
I'm ok about copyright.

awalterschulze:
I am struggling to follow what you are saying.
I think you might want to generate protobuf documentation from the comments?

ceram1:
Similar. I want to generate api documentations and server endpoints from ```.proto``` file.

PR #109, opened 2015-10-02
Casttype for maps

olegdunkan:
Hello, I have made some changes to allow to define alias for map. It is a rashly update but it works in my case. The reason was next:  
I needed the way to unmarshal json string to structures generated by gogo. But for maps like, for example, map[int]string it is necessary to implement custom marshaling, that is why aliases for map has arisen in my cases (first: type MapIntToString map[int]string, second implement interface: func (this *MapIntToString) UnmarshalJSON(b []byte) (err error) {}). Please, add quality support for casttype attribute for maps. Thanks!

### DIFF
--- a/protoc-gen-gogo/generator/generator.go
+++ b/protoc-gen-gogo/generator/generator.go
@@ -1850,8 +1850,13 @@ func (g *Generator) generateMessage(message *Descriptor) {
 				default:
 					valType = strings.TrimPrefix(valType, "*")
 				}
-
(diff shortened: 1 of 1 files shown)


awalterschulze:
Maybe add a test case that shows how it is supposed to work?

awalterschulze:
I am going to close this, since 
https://github.com/gogo/protobuf/pull/120
should be going in soon and it covers this case.

PR #119, opened 2015-11-24
Honor casttype on maps

### LINKED ISSUE #116: casttype ignored when generating maps in proto2?
The following proto2 IDL

```
// LimitRangeItem defines a min/max usage limit for any resource that matches on kind.
message LimitRangeItem {
  ...

  map<string, k8s.io.kubernetes.pkg.api.resource.Quantity> max = 2 [(gogoproto.nullable) = false, (gogoproto.customname) = "Max", (gogoproto.casttype) = "ResourceList"];

  ...
}
```

where `ResourceList` is `map[string]*resource.Quantity`.  That results in the following Go code:

```
// LimitRangeItem defines a min/max usage limit for any resource that matches on kind.
type LimitRangeItem struct {
	Max map[string]*k8s_io_kubernetes_pkg_api_resource.Quantity `protobuf:"bytes,2,rep,name=max,casttype=ResourceList" json:"max" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
}
```

with protoc 3.0.0 and gogoprotobuf 2093b57

I would assume that the map type in the Go struct should be using the alias - and that the generated unmarshaller would cast to it, but that is not the case:

```
// unmarshal
mapvalue := &k8s_io_kubernetes_pkg_api_resource.Quantity{}
if err := mapvalue.Unmarshal(data[iNdEx:postmsgIndex]); err != nil {
	return err
}
iNdEx = postmsgIndex
if m.Max == nil {
	m.Max = make(map[string]*k8s_io_kubernetes_pkg_api_resource.Quantity)
}
m.Max[mapkey] = mapvalue
iNdEx = postIndex
```

Perhaps I should not be attempting to use map with proto2 (although I noticed that there is a test for it in proto2 syntax, so I was unsure what the exact support was).

smarterclayton:
Respects the casttype extension for maps in Go.

Fixes #116

### DIFF
--- a/protoc-gen-gogo/generator/generator.go
+++ b/protoc-gen-gogo/generator/generator.go
@@ -1747,6 +1747,44 @@ func (g *Generator) GoType(message *Descriptor, field *descriptor.FieldDescripto
 	return
 }
 
+func (g *Generator) GoMapType(d *Descriptor, field *descriptor.FieldDescriptorProto) (typ string, keyTag string, valTag string) {
+	// Figure out the Go types and tags for the key and value types.
+	keyField, valField := d.Field[0], d.Field[1]
+	keyType, keyWire := g.GoType(d, keyField)
+	valType, valWire := g.GoType(d, valField)
+	keyTag, valTag = g.goTag(d, keyField, keyWire), g.goTag(d, valField, valWire)
+
+	if gogoproto.IsCastType(field) {
+		var packageName string
+		var err error
+		packageName, typ, err := getCastType(field)
+		if err != nil {
+			g.Fail(err.Error())
+		}
+		if len(packageName) > 0 {
+			g.customImports = append(g.customImports, packageName)
+		}
+
+		return typ, keyTag, valTag
+	}
+
+	// We don't use stars, except for message-typed values.
+	// Message and enum types are the only two possibly foreign types used in maps,
+	// so record their use. They are not permitted as map keys.
+	keyType = strings.TrimPrefix(keyType, "*")
+	switch *valField.Type {
+	case descriptor.FieldDescriptorProto_TYPE_ENUM:
+		valType = strings.TrimPrefix(valType, "*")
+		g.RecordTypeUse(valField.GetTypeName())
+	case descriptor.FieldDescriptorProto_TYPE_MESSAGE:
+		g.RecordTypeUse(valField.GetTypeName())
+	default:
+		valType = strings.TrimPrefix(valType, "*")
+	}
+
+	return fmt.Sprintf("map[%s]%s", keyType, valType), keyTag, valTag
+}
+
 func (g *Generator) RecordTypeUse(t string) {
 	if obj, ok := g.typeNameToObject[t]; ok {
 		// Call ObjectNamed to get the true object to record the use.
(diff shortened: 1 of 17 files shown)


smarterclayton:
Getter generation is still broken here.

awalterschulze:
Aha so instead of castkey and castvalue rather casttype the whole maptype.  That works for me.

protoc-gen-gogo/generator/generator.go
```diff
@@ -1880,27 +1918,8 @@ func (g *Generator) generateMessage(message *Descriptor) {
...
-					g.RecordTypeUse(valField.GetTypeName())
-				case descriptor.FieldDescriptorProto_TYPE_MESSAGE:
-					g.RecordTypeUse(valField.GetTypeName())
-				default:
-					valType = strings.TrimPrefix(valType, "*")
-				}
-
-				typename = fmt.Sprintf("map[%s]%s", keyType, valType)
+				var keyTag, valTag string
+				typename, keyTag, valTag = g.GoMapType(d, field)
```
awalterschulze: I would prefer not factoring it out into a function, since I have to merge in future changes from golang/protobuf.  For that reason I try to keep the diff as small as possible.
generator.go is typically the hardest to maintain, since the diff is already so large, but every little bit helps.

test/casttype/casttype.proto
```diff
@@ -62,5 +62,6 @@ message Castaway {
 	optional bytes MyBytes = 5 [(gogoproto.casttype) = "github.com/gogo/protobuf/test/casttype.Bytes"];
 	optional bytes NormalBytes = 6;
 	repeated uint64 MyUint64s = 7 [(gogoproto.casttype) = "github.com/gogo/protobuf/test/casttype.MyUint64Type"];
+	map<string,uint64> MyMap = 8 [(gogoproto.casttype) = "github.com/gogo/protobuf/test/casttype.MyMapType"];
```
awalterschulze: Can casttype of maps handle message types?

awalterschulze:
exciting :)

protoc-gen-gogo/generator/generator.go
smarterclayton: Hrm - I'm about to push the castkey and castvalue change which builds on that -  it was harder there to get the appropriate info (need a fair amount of calculation for maps) and so I have an even larger method.  Can you comment on an alternative structure there so I can get an idea what is easier to maintain for you?

test/casttype/casttype.proto
smarterclayton: Will add some tests

protoc-gen-gogo/generator/generator.go
smarterclayton: Please see #120 and https://github.com/smarterclayton/protobuf/commit/e47b2f80c09c70032dc26db00f651b5810deee36#diff-c5ab46099bd055be2629de524005c350R1762 and let me know what is going to be easier to maintain.  Should I create a new file that offers this type and leave the old factoring?

awalterschulze: I answered this in the other pull request

awalterschulze:
I think everything is in the other pull request, so I am going to close this one.
Reopen it if I am mistaken.

awalterschulze:
@smarterclayton Would you mind if I add kubernetes to list of gogoprotobuf users?

smarterclayton:
Please do

smarterclayton:
The go2idl generator is almost complete - will write something up on how it
is used and send it over so you can give it a try (it's not yet split into
a separate repo, but someone can try and consume it if they are brave in
the near term).

awalterschulze:
Thank you very much :) Kubernetes added http://gogo.github.io/

I am really excited about go2idl :)

PR #120, opened 2015-11-24
Support castkey and castvalue in maps

### LINKED ISSUE #117: Support for key and value type casttype on maps?
I would like to be able to alias key and value types on maps via `casttype`, to allow the generated Go struct to have a strongly typed key and value.  Is this something that you would accept a PR for?

For instance, turn this IDL:

```
  map<string, k8s.io.kubernetes.pkg.api.resource.Quantity> max = 2 [(gogoproto.casttype.key) = "ResourceName", (gogoproto.customname) = "Max"];
```

into Go

```
  Max map[ResourceName]resource.Quantity
```

Is there an alternative way to approach generating output like that?

smarterclayton:
Builds on #119, since there is overlap in the refactorings necessary
to get info out of the map.

Fixes #117


protoc-gen-gogo/generator/generator.go
```diff
@@ -1747,6 +1758,74 @@ func (g *Generator) GoType(message *Descriptor, field *descriptor.FieldDescripto
 	return
 }
 
+// GoMapDescriptor is a full description of the map output struct.
```
awalterschulze: Yeah I guess I'll just maintain this, since I also can't think of any easier way to do this.
Looks good :)

awalterschulze:
wow this is amazing.
Thank you so much.

awalterschulze:
Are you ok with giving over your copyright on this code?

smarterclayton:
Yes I am ok with giving over copyright.

I think the files with no changes are an artifact of making a change in the original PR that is undone by the second PR.  I can squash down once you are closer to final review.

smarterclayton:
Everything in here is in the latter PR, so I'll close this out and we can squash that one down to the correct, isolated commits when the review is finished.

awalterschulze:
Perfect, except you don't need to squash.
I might even prefer the documented process (unsquashed), rather than the documented hindsight (squashed).
As long as the last commit that we merge into master makes everything green again and all the issues are being addressed (as they are) I am happy.

PR #121, opened 2015-11-26
Support nullable=false on a map to mean embedded map value

### LINKED ISSUE #116: casttype ignored when generating maps in proto2?

### LINKED ISSUE #118: Support for nullable = false map values?
The following IDL

```
message LimitRangeItem {
  map<string, k8s.io.kubernetes.pkg.api.resource.Quantity> max = 2 [(gogoproto.casttype) = "ResourceList", (gogoproto.nullable) = false, (gogoproto.customname) = "Max"];
}
```

generates the following Go struct

```
// LimitRangeItem defines a min/max usage limit for any resource that matches on kind.
type LimitRangeItem struct {
    Max map[string]*k8s_io_kubernetes_pkg_api_resource.Quantity `protobuf:"bytes,2,rep,name=max,casttype=ResourceList" json:"max" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
}
```

But I would prefer the value field to be not nullable.  Since Go maps are always references, would you accept a PR that uses nullable to guide the generation to treat the value as an embedded type, not a pointer?

smarterclayton:
Fixes #116, #117, #118

Builds upon #119 and #120


test/casttype/combos/both/casttype.proto
```diff
@@ -62,5 +62,11 @@ message Castaway {
 	optional bytes MyBytes = 5 [(gogoproto.casttype) = "github.com/gogo/protobuf/test/casttype.Bytes"];
 	optional bytes NormalBytes = 6;
 	repeated uint64 MyUint64s = 7 [(gogoproto.casttype) = "github.com/gogo/protobuf/test/casttype.MyUint64Type"];
+	map<string,uint64> MyMap = 8 [(gogoproto.casttype) = "github.com/gogo/protobuf/test/casttype.MyMapType"];
+	map<string,uint64> MyCustomMap = 9 [(gogoproto.castkey) = "github.com/gogo/protobuf/test/casttype.MyStringType", (gogoproto.castvalue) = "github.com/gogo/protobuf/test/casttype.MyUint64Type"];
+  map<int32,Wilson> MyMessageMap = 10 [(gogoproto.castkey) = "github.com/gogo/protobuf/test/casttype.MyInt32Type", (gogoproto.nullable) = false];
```
awalterschulze: We could probably have two Wilson tests, one with nullable=false and one without?

test/mapsproto2/combos/both/mapsproto2.pb.go
```diff
@@ -2554,7 +2564,7 @@ func NewPopulatedAllMaps(r randyMapsproto2, easy bool) *AllMaps {
 		v2 := r.Intn(10)
 		this.StringToDoubleMap = make(map[string]float64)
 		for i := 0; i < v2; i++ {
-			v3 := randStringMapsproto2(r)
+			v3 := string(randStringMapsproto2(r))
```
awalterschulze: These casts are a little redundant aren't they?

test/mapsproto2/combos/both/mapsproto2.pb.go
```diff
@@ -2993,9 +3003,9 @@ func (this *AllMaps) String() string {
 		keysForStringToFloatMap = append(keysForStringToFloatMap, k)
 	}
 	github_com_gogo_protobuf_sortkeys.Strings(keysForStringToFloatMap)
-	mapStringForStringToFloatMap := "map[string]*float32{"
+	mapStringForStringToFloatMap := "map[string]float32{"
 	for _, k := range keysForStringToFloatMap {
-		mapStringForStringToFloatMap += fmt.Sprintf("%v: %v,", k, this.StringToFloatMap[k])
+		mapStringForStringToFloatMap += fmt.Sprintf("%#v: %#v,", k, this.StringToFloatMap[k])
```
awalterschulze: %v calls String and %#v calls GoString
We want the String method to recursively call String and not GoString

awalterschulze:
Almost there :)
This is really amazing work.
Thank you so much.

test/mapsproto2/combos/both/mapsproto2.pb.go
smarterclayton: Hrm, did not intend this change.  Also, I need to check on the float32 pointer - that's also not expected.

test/casttype/combos/both/casttype.proto
smarterclayton: Yes

gogoproto/doc.go
```diff
@@ -45,7 +45,9 @@ Gogoprotobuf tries to fix these problems with the nullable, embed, customtype an
   - embed, if true, the field is generated as an embedded field.
   - customtype, It works with the Marshal and Unmarshal methods, to allow you to have your own types in your struct, but marshal to bytes. For example, custom.Uuid or custom.Fixed128
   - customname (beta), Changes the generated fieldname. This is especially useful when generated methods conflict with fieldnames.
-  - casttype (beta), Changes the generated fieldtype.  All generated code assumes that this type is castable to the protocol buffer field type.  It does not work for maps, structs or enums.
+  - casttype (beta), Changes the generated fieldtype.  All generated code assumes that this type is castable to the protocol buffer field type.  It does not work for structs or enums.
+  - castkey (beta), Changes the generated fieldtype for a map key.  All generated code assumes that this type is castable to the protocol buffer field type.  Only supported on maps
+  - castvalue (beta), Changes the generated fieldtype for a map value.  All generated code assumes that this type is castable to the protocol buffer field type.  Only supported on maps.
```
awalterschulze: We should probably mention that this is not supported on message types?

plugin/unmarshal/unmarshal.go
```diff
@@ -675,26 +677,60 @@ func (p *unmarshal) field(file *descriptor.FileDescriptorProto, msg *generator.D
 			p.P(`}`)
 			p.P(`m.`, fieldname, ` = &`, p.OneOfTypeName(msg, field), `{v}`)
 		} else if generator.IsMap(file, field) {
-			mapMsg := generator.GetMap(file, field)
-			keyField, valueField := mapMsg.GetMapFields()
+			m := p.GoMapType(nil, field)
+			mapgoTyp, keyField, valueField, keyAliasField, valueAliasField := m.GoType, m.KeyField, m.ValueField, m.KeyAliasField, m.ValueAliasField
```
awalterschulze: This is a bit nit picky, but it seems like it is possible to just use m.KeyField, etc rather than assign them to other variables that basically have the same name.  What do you think?

protoc-gen-gogo/generator/helper.go
```diff
@@ -220,6 +220,55 @@ func IsMap(file *descriptor.FileDescriptorProto, field *descriptor.FieldDescript
...
+
+	nullable := gogoproto.IsNullable(field)
+	if err := proto.SetExtension(valField.Options, gogoproto.E_Nullable, &nullable); err != nil {
+		g.Fail(err.Error())
+	}
+	return valField
+}
+
+// cloneFieldExtensions creates a copy of field where field.Options.XXX_extensions is safe to mutate.
+func cloneFieldExtensions(field *descriptor.FieldDescriptorProto) *descriptor.FieldDescriptorProto {
```
awalterschulze: Have you tried the proto.Clone function?
It should work for the FieldDescriptor, since it has not been modified by nullable=false and customtypes etc.

awalterschulze:
I am only nit picking now, because everything looks correct and pretty much ready to be merged.

I also think I need to ask again, since this is a new pull request.
Are you ok with giving over your copyright on the code in pull request 121?

awalterschulze:
Are you ready for me to merge it?

smarterclayton:
I'll go through and address those comments - just thinking of the next person to come in here :)  Yes, I am fine with transferring copyright for this pull request.

gogoproto/doc.go
smarterclayton: Hrm - I can add a test, I thought it would be supported.  Let me try that.

protoc-gen-gogo/generator/helper.go
smarterclayton: Hrm, proto.Clone fails to copy the FieldOptions struct.  Weird.

smarterclayton: Nm, forgot that I initialized it myself.

gogoproto/doc.go
smarterclayton: I forgot, this is hard to test with the structure of test cast type.  It will work when the alias type is in the same package and points to the generated struct (we have an example of it in Kube).  Will think on how to deliver a better test here - is a follow up fine?

smarterclayton:
Fixed more edge cases and applied review feedback.  See the jsonpb commit - I'm not sure this is the best solution, but without access to the proto.Properties map key and value data, it's a harder problem.

gogoproto/doc.go
awalterschulze: Yes I think just say we don't support messages for castvalue.
If you can figure out how to support and test it, then we just update the doc at that point.

plugin/stringer/stringer.go
```diff
@@ -150,25 +150,35 @@ func (p *stringer) Generate(file *generator.FileDescriptor) {
 				continue
 			}
 			fieldname := p.GetFieldName(message, field)
-			mapMsg := generator.GetMap(file.FileDescriptorProto, field)
-			keyField, valueField := mapMsg.GetMapFields()
+
+			m := p.GoMapType(nil, field)
+			mapgoTyp, keyField, keyAliasField := m.GoType, m.KeyField, m.KeyAliasField
```
awalterschulze: We can just reuse the values again instead of assigning new ones.

awalterschulze:
Maybe we can get to a point where we can merge before supporting castvalue of message types.
And then think about casttype and castvalue for message types, what do you think?

smarterclayton:
Well, I believe support is complete for that - it's just whether the jsonpb
changes worry you or not.

awalterschulze:
Unfortunately they do worry me if they are there and not properly tested.
Lets rather drop the feature until it can properly be supported?

The other thing is that the tests are breaking for protoc 2.5 and 2.6.1, see the travis build.
I have made another comment about how to fix that.

smarterclayton:
By not properly tested - I believe the castvalue test case exercises all the code paths (but I could be wrong).  The problem of castvalue I believe is a general problem, not specific to nullable message vs non-nullable message.

The use case we had was for map[string]MessageAlias - so I need to solve castvalue of a message one way or another.

smarterclayton:
The proto error was fixed - castvalue copied and pasted something that did not support maps.

awalterschulze:
I think castvalue and casttype for message types are very hard.
I think we could merge everything without the changes made for castvalue message types and then try to tackle that separately, just because I think it is a hard problem.
If you want to continue in this pull request, you are welcome.
I just think splitting the work here will make this a little easier.

smarterclayton:
I may be missing the areas that you're most concerned with.  Is there generated code that combo doesn't touch that I may be missing?  

Two more commits added that cover castvalue more thoroughly (with fixes)

awalterschulze:
Sorry about the delay.
I was on holiday yesterday.

awalterschulze:
I can't remember, but I think you mentioned a problem with imported casttype messages?
I think the generated code will cause a circular dependancy.  So its a easy case not to support even though messages that are in the same package actually do seem to work.

plugin/equal/equal.go
```diff
@@ -368,10 +368,33 @@ func (p *plugin) generateField(file *generator.FileDescriptor, message *generato
...
+				m := p.GoMapType(nil, field)
+				valuegoTyp, _ := p.GoType(nil, m.ValueField)
+				valuegoAliasTyp, _ := p.GoType(nil, m.ValueAliasField)
+				nullable, valuegoTyp, valuegoAliasTyp = generator.GoMapValueTypes(field, m.ValueField, valuegoTyp, valuegoAliasTyp)
+
+				mapValue := m.ValueAliasField
 				if mapValue.IsMessage() || p.IsGroup(mapValue) {
-					p.P(`if !this.`, fieldname, `[i].Equal(that1.`, fieldname, `[i]) {`)
+					switch {
+					case nullable && valuegoTyp == valuegoAliasTyp:
```
awalterschulze: This could just be a normal if and else?

awalterschulze:
Thank you very much smartest clayton.
This is some excellent work.  I really appreciate your efforts.

plugin/equal/equal.go
awalterschulze: Don't worry I fixed this.

PR #136, opened 2016-01-15
Don't gofmt non-Go files.

fd:


awalterschulze:
Interesting what json files are you generating?

fd:
For now I'm just generating Swagger specifications for my GRPC/HTTP gateway.
But other things will follow. (things like: authz policies, etc...)

awalterschulze:
Wow have you open sourced it?

fd:
No not yet. It's really a toy implementation for now. But open sourcing the generator is definitely part of the plan.

awalterschulze:
Why do you need to do this through the vanity binary?
It seems quite nice

fd:
Well, I guess I don't have todo through the vanity package. Right now I'm just appending some files to the protoc response from a custom plugin.

awalterschulze:
Sorry I have been sick for a while.
Have you made any progress without using the vanity package?

fd:
No I haven't. I'm not sure how to make a non-vanity plugin.

awalterschulze:
Basically you do exactly what you did in the diff here, but you just copy it into your own library and create your own binary.  The only reason I am suggesting this is because I think its niche enough to justify a copy.

awalterschulze:
I suspect I can close this.

PR #139, opened 2016-01-19
add float32/64 casttype tests

tamird:
This is more bug report than PR since tests don't pass. @awalterschulze could you have a look?


awalterschulze:
Sorry for the wait. I have been sick.
I have a fix can we have the copyright of for the test you have created?

tamird:
Yep

awalterschulze:
Thanks for reporting :)  I have pushed the fix.

PR #140, opened 2016-01-24
Extension: generate ProtoSize method

### LINKED ISSUE #56: Bug: Conflict with Size fieldname and Size method name
This issue was originally reported here: https://github.com/gogo/protobuf/issues/39.

I see that when you have a field named String golang/protobuf generates that fieldname as String_, since this field conflict with the generated String method.

I think this would be a great solution to the Size method and fieldname conflict problem.

dennwc:
Implemented alternative extension to generate ProtoSize methods instead of Size. This allows to use Size as a field name in generated messages.

Fixes #56

Large diff is due to full regeneration of files.


plugin/protosizer/size.go
```diff
@@ -0,0 +1,590 @@
+// Copyright (c) 2013, Vastech SA (PTY) LTD. All rights reserved.
```
awalterschulze: You could probably reuse the code in plugin/size/size.go to generate both Size and ProtoSize methods, just like Equal and VerboseEqual methods are both generated by the equal plugin and marshaler and unsafemarshaler and both generated by the marshaler plugin.

protoc-gen-gogo/generator/generator.go
```diff
@@ -1865,7 +1864,7 @@ var methodNames = [...]string{
 }
 
 // Generate the type and default constant definitions for this Descriptor.
-func (g *Generator) generateMessage(message *Descriptor) {
+func (g *Generator) generateMessage(file *FileDescriptor, message *Descriptor) {
```
awalterschulze: Are you sure we can't get the FileDescriptor given the Generator and the Descriptor.
Would be nice if we don't need to change this function signature.

protoc-gen-gogo/generator/helper.go
```diff
@@ -181,10 +181,19 @@ func (g *Generator) GetFieldName(message *Descriptor, field *descriptor.FieldDes
 			return fieldname + "_"
 		}
 	}
+	if gogoproto.IsProtoSizer(file.FileDescriptorProto, message.DescriptorProto) {
```
awalterschulze: The reserved word list should take of this.  This code should not be necessary.

test/protosize/protosize.proto
```diff
@@ -0,0 +1,17 @@
...
+
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.protosizer_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.populate_all) = true;
+option (gogoproto.testgen_all) = true;
+option (gogoproto.equal_all) = true;
+
+message SizeMessage {
+	optional int64 size = 1;
```
awalterschulze: Please also include a ProtoSize field

awalterschulze:
Looks very promising :)
Thanks for the hard work.

protoc-gen-gogo/generator/helper.go
dennwc: But I can't add both Size and ProtoSize to reserved names - it will try to put suffix on them always. Or I missed something?

awalterschulze: No my bad.  Maybe add ProtoSize as a reserved name, but yes we are trying to make Size not a reserved word.  So keep the check only for the Size method.

dennwc:
@awalterschulze Looks like everything is done. Do I need to squash commits?

awalterschulze:
No need to squash commits.
Everything looks great :)

dennwc:
Yes, that's ok

awalterschulze:
Thank you very much for this work.

awalterschulze:
@msolo-dropbox you might want to take a look at this.
Hopefully this solves your reason for a fork.

PR #143, opened 2016-02-01
Fix: Correct proto.Size invocation for tests

dennwc:
Protosizer was used to generate wrong (`proto.ProtoSize` instead of `proto.Size`) invocations for tests.
Missed this because global Makefile was not updated properly.


awalterschulze:

dennwc:
Yep

awalterschulze:
Thanks again.
I missed it too :)

PR #148, opened 2016-02-19
gogoproto: enum and enum value custom name extension

### LINKED ISSUE #94: Vote For New Feature: customname for enum value.

stevvooe:
This provides support for custom enum and enum value names allowing one
to override default naming to their delight. The `enum_customname`
extension will allow one to set the type name of an enum. If
`goproto_enum_prefix` is enabled, this value will be used as a prefix
when generating enum values. If `enumvalue_customname` is enabled for a
particular value, the prefix is ignored and the name is used directly.

A new extension set is declared for `EnumValueOptions` in the 66xxx
range to continue with the extension numbering convention.

Signed-off-by: Stephen J Day <stephen.day@docker.com>

This is the first movement in the plan outlined in #149.

Closes #94.


protoc-gen-gogo/generator/generator.go
```diff
@@ -1428,6 +1428,10 @@ func (g *Generator) generateEnum(enum *EnumDescriptor) {
 		g.PrintComments(fmt.Sprintf("%s,%d,%d", enum.path, enumValuePath, i))
 
 		name := ccPrefix + *e.Name
+		if gogoproto.IsEnumCustomName(e) {
```
awalterschulze: The plugin should probably check that EnabledGoEnumPrefix is false if IsEnumCustomName is true.
If this is not the case the plugin should at least give a WARNING and possibly an error.
See the defaultcheck plugin for examples.

gogoproto/gogo.proto
```diff
@@ -38,6 +38,10 @@ extend google.protobuf.EnumOptions {
 	optional bool enum_stringer = 62022;
 }
 
+extend google.protobuf.EnumValueOptions {
+	optional string enum_customname = 66001;
```
awalterschulze: We could probably make it 65001, since that is the next range.

awalterschulze:
This seems good sofar, but I can not merge this pull request without it having enough tests.
Do you know how to create tests?

Basically create a new folder in the test folder.
See all the other folders as examples.
Create a proto file which uses your new extension.
Try and combine this in various ways with other features.
See if anything breaks.

gogoproto/gogo.proto
stevvooe: 65xxx seems to be in use for field options: https://github.com/gogo/protobuf/blob/master/gogoproto/gogo.proto#L102.

protoc-gen-gogo/generator/generator.go
stevvooe: I'll try to match the existing precedence choices if possible. As a user, I would think the more specific customname (enum value) would always override a less-specific (enum) prefix. I'll make sure there is a test case for this.

stevvooe:
@awalterschulze I wanted to make sure the code looked good before picking up the test system. Sounds like we're good to go. I'll update this PR with the tests in the next few days.

protoc-gen-gogo/generator/generator.go
awalterschulze: I prefer an error or warning.  Especially since we are going to stealthily apply this extension.  The user should know when he is specifying an extension that will have no effect.

gogoproto/gogo.proto
awalterschulze: my bad :)

awalterschulze:
Yes we are good to go :)

protoc-gen-gogo/generator/generator.go
stevvooe: @awalterschulze Looks like `goproto_enum_prefix` is enabled by default. It wouldn't make much sense to emit a warning or error if `enum_customname` is also enabled since this would be triggered with common usage. I can still add it if you insist.

stevvooe:
@awalterschulze PR updated with tests. Note that I've separated this out to allow one to control both the name of the enum as well as its values.

awalterschulze:
I see now how the prefix and customname works together.
I thought the customname was going to overwrite the whole name, but this only happens if the prefix is also set to false.
That could work, but is that what you intended?

test/enumcustomname/enumcustomname.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto2";
+package enumcustomname;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+enum MyEnum {
+	option (gogoproto.goproto_enum_stringer) = true; // ensure it behaves correctly with stringer.
```
awalterschulze: goproto_enum_stringer is true by default

awalterschulze:
This is a good start :)

I think there should at least one more test where these enums are used inside messages.  This should make sure that the enum name is always calculated in the right way and not accidentally by some other shortcut.

Also see my comment on goproto_enum_stringer, this should tested with false.
Also gogoproto.enum_stringer should be tested with true.

But it seems this feature might be easier that I thought.

test/enumcustomname/enumcustomname.proto
stevvooe: I'll move that down to the disabled prefix test, since we really want to make sure the stringer is being generated with the overrides.

stevvooe:
> I thought the customname was going to overwrite the whole name, but this only happens if the prefix is also set to false.
That could work, but is that what you intended?

When doing the testing, I found that we really needed an enum type name (`enum_customname`) and enum value name extension (`enumvalue_customname`). The type clarified things around the use of prefixes. The approach now is fairly clear.

I discovered a few problems with the use as fields, so that was a worthwhile exercise. I've also added some tests for cross-package import and default values. I'll push those shortly.

stevvooe:
@awalterschulze Alright, updates are pushed.

Note that I also removed the separate test package for `enumvalue_customname`, as it made more sense to test it together with `enum_customname`.

test/thetest.proto
```diff
@@ -386,6 +386,22 @@ message AnotherNinOptEnumDefault {
 	optional AnotherTestEnum Field1 = 1 [default=E];
 }
 
+// YetAnotherTestEnum is used to test cross-package import of custom name
```
awalterschulze: If you are going to declare these, you should probably also use them in a message, that way you can get some free combo testing :)

Makefile
```diff
@@ -69,6 +69,8 @@ regenerate:
 	make -C test/moredefaults regenerate
 	make -C test/issue8 regenerate
 	make -C test/enumprefix regenerate
+	make -C test/enumcustomname regenerate
```
awalterschulze: I guess you only need the one know?

test/thetest.proto
stevvooe: Done.

Makefile
stevvooe: Silly.

test/thetest.proto
```diff
@@ -348,44 +348,73 @@ enum TheTestEnum {
...
 	optional AnotherTestEnum Field1 = 1;
+	optional YetAnotherTestEnum Field2 = 2;
+	optional YetYetAnotherTestEnum Field3 = 3;
 }
 
 message AnotherNinOptEnumDefault {
 	option (gogoproto.goproto_getters) = true;
 	option (gogoproto.face) = false;
 	optional AnotherTestEnum Field1 = 1 [default=E];
+	optional YetAnotherTestEnum Field2 = 2 [default=BB];
```
awalterschulze: Yes! I like all of these tests :D

protoc-gen-gogo/generator/generator.go
```diff
@@ -2178,6 +2194,31 @@ func (g *Generator) generateMessage(message *Descriptor) {
...
+			} else {
+				// This condition should be rare or impossible. Could be caused
+				// by a compiler bug that didn't ensure that default enum value
+				// is define for the type.
+				g.Fail(fmt.Sprintf("could not resolve default enum value for %v.%v",
+					g.DefaultPackageName(obj), def))
+
+			}
+
 			if gogoproto.EnabledGoEnumPrefix(enum.file, enum.EnumDescriptorProto) {
```
awalterschulze: Maybe this if should be inside the enum prefix method?
I know its my bad code, but could you try to move it and see if its better?

protoc-gen-gogo/generator/generator.go
```diff
@@ -2178,6 +2194,31 @@ func (g *Generator) generateMessage(message *Descriptor) {
 				log.Printf("don't know how to generate constant for %s", fieldname)
 				continue
 			}
+
+			// hunt down the actual enum corresponding to the default
+			var enumValue *descriptor.EnumValueDescriptorProto
+			for _, ev := range enum.Value {
+				// TODO: This could probably be preindexed but the performance
```
awalterschulze: Remove this comment.

protoc-gen-gogo/generator/generator.go
```diff
@@ -2178,6 +2194,31 @@ func (g *Generator) generateMessage(message *Descriptor) {
...
+					enumValue = ev
+				}
+			}
+
+			if enumValue != nil {
+				if gogoproto.IsEnumValueCustomName(enumValue) {
+					def = gogoproto.GetEnumValueCustomName(enumValue)
+				}
+			} else {
+				// This condition should be rare or impossible. Could be caused
```
awalterschulze: Remove this comment, the error explains it well enough.

awalterschulze:
LGTM :)

protoc-gen-gogo/generator/generator.go
stevvooe: I actually gave this a try but ran into some odd consistency issues. 

Basically, the complexity here is that a Go namespace and a Protobuf namespace needs to be maintained. It is not wildly complex but would have made this PR much more complex. An example of the problem is that enum value search loop at line 2200. Really, each type needs to produce its Go name and Protobuf name, qualified to the file scope. This led me down the path of an interface change, which is much too big of a change when adding a small feature.

If you don't mind, I'd prefer to leave that to a refactoring, which I might look at undertaking if I do more work in here.

test/thetest.proto
stevvooe: They actually caught a few subtle bugs!

stevvooe:
@awalterschulze I removed some of the comments and rebased with master. We should be good to go!

awalterschulze:
Thank you for your work.
Are you ok with giving over your copyright?

awalterschulze:
Also lets just make sure we have the logic right, since the code looks perfect.
The gostyle plugin is it going to keep enum_prefix set to true?

awalterschulze:
Sorry for taking so long to reply.

stevvooe:
> Thank you for your work.
Are you ok with giving over your copyright?

Yes, as long as there is authorship credit. This contribution is on the behalf of Docker, but we allow assignment when contributing upstream.

> The gostyle plugin is it going to keep enum_prefix set to true?
And will that pass golint?

Interesting you ask. There are a few options here. With the way it is now, prefix+customname will generate `Prefix_Custom` pairs, if enabled. This retains compatibility with the current generation approach. For `gostyle`, we can either disable the `prefix` extension and generate all the custom names directly or have another extension `prefix_nounderscore` which removes the underscore.

awalterschulze:
What do mean with an authorship credit? sorry I just haven't encountered it before.  Would you like to add your name to the contributors file, if so please add it to the pull request.

I would like to avoid adding another extension.
I think I like your plan of gostyle setting enumprefix to false and then doing all the work in customname.
On the other hand if customname took over the whole name by default then enumprefix could also be implemented using customname the same way we are going to implement gostyle using customname.

Is docker using gogoprotobuf?
Can I add them to the user list on gogo.github.io?

stevvooe:
> What do mean with an authorship credit? sorry I just haven't encountered it before. Would you like to add your name to the contributors file, if so please add it to the pull request.

That is all that is required. Added. ;)

You may want to automate CONTRIBUTORS generation. We do something like this:

```console
$ git log --format='%aN <%aE>' | sort -fu                                                                                                                                                                                            
Anton Povarov <anton.povarov@gmail.com>
Anton Povarov <antoxa@corp.badoo.com>
awalterschulze <awalterschulze@gmail.com>
awalterschulze@gmail.com <awalterschulze@gmail.com>
ceram1 <ceram1000@gmail.com>
Clayton Coleman <ccoleman@redhat.com>
dennwc <denis.smirnov.91@gmail.com>
Denys Smirnov <denis.smirnov.91@gmail.com>
Dwayne Schultz <dschultz@pivotal.io>
Georg Apitz <gapitz@pivotal.io>
Gustav Paul <gustav.paul@gmail.com>
John Tuley <john@tuley.org>
John Tuley <jtuley@pivotal.io>
Laurent <laurent@adyoulike.com>
Patrick Lee <patrick@dropbox.com>
Stephen J Day <stephen.day@docker.com>
Tamir Duberstein <tamird@gmail.com>
Todd Eisenberger <teisenberger@dropbox.com>
Tormod Erevik Lea <tormodlea@gmail.com>
Walter Schulze <awalterschulze@gmail.com>
Walter Schulze <walter@vastech.co.za>
```

You can remove the duplicates by creating a `.mailmap` file.

> I would like to avoid adding another extension.
I think I like your plan of gostyle setting enumprefix to false and then doing all the work in customname.
On the other hand if customname took over the whole name by default then enumprefix could also be implemented using customname the same way we are going to implement gostyle using customname.

Those are the current options. I think agree that deferring to `customname` is correct but let's take that discussion over to #149.

awalterschulze:
Thanks for the contributors tip.  I added it :)
https://github.com/gogo/protobuf/commit/080283ddb9619deeef8cf059c79851e6f1966c07

stevvooe:
@awalterschulze No problem. Great to see this merged!

awalterschulze:
Oh yes and thank you for your excellent work here :)

Issue #149, opened 2016-02-19
gostyle: provide option to generate golint compatible protobufs


stevvooe:
Many users of protobuf would prefer to directly use generated structures within their application. The main benefit of using the generated types directly is reduced copying, allocation and bugs when marshaling and unmarshaling types. The prevailing issue is that generated protobuf code does not pass `golint`. Such types also don't work well with godoc documentation.

We experimented with this concept applied to enums in #133. The results were encouraging but others felt that we should provide this option for other types, at the file level.

There are several steps that need to happen to provide this feature.

Here is current plan:

- [ ] Implement `enum_customname` extension to allow one to provide custom names for enum values. This will address #94. `gostyle` transforms will leverage this across the board. (#148)
- [ ] Implement `gostyle` extension
  - [ ] Add a "transform" step to allow generators to make a pass over the descriptor to apply extensions before generation.
  - [ ] Define variable name transforms that are compatible with golint (see https://github.com/golang/lint/blob/master/lint.go#L510 for information on this transform).
  - [ ] Implement modifications to customname entries in the pre-generation descriptor pass.

It has been pointed out that there may be unforeseen complexities in this endeavor. If the plan changes or we need more steps, I'll try to update this issue accordingly.

If anyone wants to help in this effort, please coordinate here.

awalterschulze:
I have also only recently seen the problems with godoc at the package level.   I don't really want the godoc reader to know that this is protocol buffer package, only that there exists protocol buffer messages within the package.  I will investigate this a little soon, hopefully next week.

customnames might also be required for messages.
customnames for fields is luckily already complete.

awalterschulze:
Ok so I would like to merge #148 , but we should discuss the choice of whether to 1) have gostyle set enum_prefix to false and have enum_customname  be used to change the field name to something appropriate, whether this included the prefix is also up for discussion, 2) have enum_customname be a conflicting extension with enum_prefix (you can't have both at the same time) and give enum_customname full control over the whole field name.  This way we could even implement enum_prefix with enum_customname and some vanity magic.

Currently #148 implements option 1.  Is this what we want in the long term?
I liked option 2 the most, but I think its extra work for very little gain.
And if we choose 1 we could still leave it up to the user to decide whether the person wants a prefix or not.
So in the end I like option 1 the most, since the user gains extra flexibility.
If you agree. I'll merge #148

stevvooe:
I agree with your assessment. 

[...]

PR #152, opened 2016-03-01
Use result of computation for ineffassign

tschottdorf:
This is a no-op semantically; the `ineffassign` lint will complain when using maps
otherwise.

### DIFF
--- a/plugin/unmarshal/unmarshal.go
+++ b/plugin/unmarshal/unmarshal.go
@@ -236,6 +236,7 @@ func (p *unmarshal) decodeVarint(varName string, typName string) {
 	p.P(`}`)
 	p.Out()
 	p.P(`}`)
+	p.P(`_ = `, varName) // satisfy ineffassign
 }
 
 func (p *unmarshal) decodeFixed32(varName string, typeName string) {
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@
 
 .PHONY: nuke regenerate tests clean install gofmt vet contributors
 
-all: clean install regenerate install tests errcheck vet
+all: clean install regenerate install tests errcheck vet ineffassign
 
 buildserverall: clean install regenerate install tests vet
 
@@ -98,6 +98,10 @@ regenerate:
 tests:
 	go build ./test/enumprefix
 	go test ./...
+
+ineffassign:
+	go get -u github.com/gordonklaus/ineffassign
+	ineffassign .
 
 vet:
 	go vet ./...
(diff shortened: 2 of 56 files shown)


awalterschulze:
Could you regenerate the code, by running make?

Why is ineffassign triggered?
Maybe this will be more clear once you have regenerated the code.

tschottdorf:
I'll wait for @tamird's PR which updates the code (and removes a bunch of apparently old `.pb.go` files). With this change, `ineffassign` passes. Essentially many calls to `decodeVarint` don't use the variable after, which is linted against.

awalterschulze:
Aha, ok I look forward to the updated code.
Please make a post here when you are ready.

Also are you ok with giving over your copyright for this pull request?

tschottdorf:
> Also are you ok with giving over your copyright for this pull request?
sure.

Code is updated, PTAL.

plugin/unmarshal/unmarshal.go
```diff
@@ -236,6 +236,7 @@ func (p *unmarshal) decodeVarint(varName string, typName string) {
 	p.P(`}`)
 	p.Out()
 	p.P(`}`)
+	p.P(`_ = `, varName) // satisfy ineffassign
```
tamird: looking at the diff in generated code, this seems to change a lot more than necessary to fix the ineffassign check. is it possible to limit this codegen change to just the map type?

tschottdorf: Maybe, but it's generated code. I had this in two locations in maps initially but it left a lot of other places. Not worth optimizing imo.

awalterschulze:
Almost every decodeVarint might be a little excessive for a check that is not even in go vet.  Maybe it would be better to find the ineffective assigns and rather turn them into skipVarints, which could possibly add a little speed?  Or maybe I am totally missing the point?

awalterschulze:
Sorry click the wrong comment button, my bad.  Its open again.

tschottdorf:
No, that's certainly possible - but it exceeds my personal commitment to this change. Happy to close this (and just ignore the lint errors in the generated gogoproto code) if you think this solution is too rough. As stated above, it's auto-generated code which generates no run-time overhead, so sitting down and manually seeking out the 100+ call sites which generate this error does not seem like a valuable use of my time.

awalterschulze:
fair enough :) , but I think I'll rather wait until https://github.com/gogo/protobuf/issues/149 is further down its path, because then I guess someone else might care enough.

PR #156, opened 2016-03-16
Add stable_marshaler option to marshal deterministically

tamird:
This is needed to support exotic use cases where deterministic
marshalling is needed.

See https://github.com/cockroachdb/cockroach/issues/5090.

@awalterschulze I'm happy to bike shed on the option name - maybe something like "deterministicmarshal" would be better and more general? Your call.

cc @vivekmenezes @tschottdorf


awalterschulze:
I like stablemarshal :)

I don't know if it needed as a field option though?
I would think your message would generate a MarshalTo method that is either stable or not.
I don't see a more find grained use case, but what do you think?  

Also I think the only thing that is causing the unstable marshal and that is maps, or am I missing something?

awalterschulze:
Also if we go the message option route then sortmarshaler is probably more consistent.

tamird:
Yep, it's only maps. Changed to a message-level option as you suggested. PTAL

awalterschulze:
LGTM.
Do you know why are the travis tests not running?

tamird:
No idea :(

awalterschulze:
Ok it looks like they ran on travis, but they just are not displaying here anymore

awalterschulze:
Its probably a small github bug with the new review look.  They'll probably fix it soon enough.

tamird:
Cool beans. You can have my copyright, as usual.

awalterschulze:

tamird:
Yep.

awalterschulze:
Thank you for your work :)

tamird:
Thanks!

awalterschulze:
I think this was probably a record for our most efficient interaction :)

tamird:
Agreed!

PR #165, opened 2016-04-04
proto3: Avoid marshalling nil customtype bytes fields

tamird:
This fixes a bug that causes protos to change after marshalling and
unmarshalling.

@awalterschulze

### DIFF
--- a/test/theproto3/Makefile
+++ b/test/theproto3/Makefile
@@ -32,3 +32,4 @@ regenerate:
 	cat footer.proto >> theproto3.proto
 	go install github.com/gogo/protobuf/protoc-gen-combo
 	protoc-gen-combo --version="3.0.0" --proto_path=../../../../../:../../protobuf/:. theproto3.proto
+	find combos -type d -not -name combos -exec cp proto3_test.go.in {}/proto3_test.go \;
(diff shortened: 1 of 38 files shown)


test/theproto3/Makefile
```diff
@@ -31,4 +31,5 @@ regenerate:
 	cat maps.proto >> theproto3.proto
 	cat footer.proto >> theproto3.proto
 	go install github.com/gogo/protobuf/protoc-gen-combo
-	protoc-gen-combo --version="3.0.0" --proto_path=../../../../../:../../protobuf/:. theproto3.proto
\ No newline at end of file
+	protoc-gen-combo --version="3.0.0" --proto_path=../../../../../:../../protobuf/:. theproto3.proto
+	find combos -type d -not -name combos -exec cp proto3_test.go.in {}/proto3_test.go \;
```
awalterschulze: Ok this looks cool and is the correct way to do it, but not everyone knows `find` that well.
Could we rather just "manually" copy the proto3_test.go.in to each folder?

awalterschulze:
The rest all looks great :)

test/theproto3/Makefile
tamird: We could, but the point was not to duplicate the directory names here. Otherwise, we may as well remove this entirely and manually copy things around =/

tamird:
Updated this slightly to use a shared helper that's easier to read. Yes, I am ok with giving over my copyright.

test/theproto3/Makefile
awalterschulze: I know, but my point is readability, but anyway I guess thats just nitpicky.

awalterschulze:
Thank you :)

tamird:
Thanks!

PR #179, opened 2016-05-13
Extra struct content

AudriusButkevicius:


test/extrastructcontent/extrastructcontent.proto
```diff
@@ -0,0 +1,15 @@
...
+package extrastructcontent;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.sizer_all) = true;
+
+message ExtraStructContent {
+	option (gogoproto.extra_struct_content) = "MyMagicValue int64\nMyOtherMagicValue string";
```
awalterschulze: Hmmm, looks good, but is there some way we could have a raw string?
```
option (gogoproto.extra_struct_content) = `
MyMagicValue int64
MyOtherMagicValue string
`;
```

awalterschulze:
Looks really good :)

test/extrastructcontent/extrastructcontent.proto
AudriusButkevicius: It does not seem the protobuf spec supports that.

stevvooe:
Rather than taking a raw Go string, would it be better to take a structured option? Check out [https://github.com/googleapis/googleapis/blob/master/google/api/annotations.proto#L26] for an example and the syntax. Basically, you would have an inline structure with a list of types and names for the extra fields. It would probably be a little more robust and nicer to read.

AudriusButkevicius:
Sadly it's above my understanding of proto syntax, so you either need to provide better examples or walk me there.

But before we do that, I'd need an acknowledgement that this new version would be merged upstream as a justification for me to spend the effort. Current version already works for my needs, and the only benefits I could reap from the new version would be not having to maintain a fork.

awalterschulze:
I'll say it definitely increases the chances. But I can't make a guarantee
and I still want at least another user

stevvooe:
@AudriusButkevicius Here is an [example](https://github.com/googleapis/googleapis/blob/master/google/cloud/speech/v1/cloud_speech.proto#L34) of the syntax. I would hope you would take the time to explore other approaches before pushing for merging this. The current suggestion just seems very fragile.

awalterschulze:
I am guessing you found another solution.  I am closing this, but please comment if want to continue work.

PR #183, opened 2016-05-31
jsonpb: Adds support for other []byte types to be base64 encoded.

d4l3k:
The proto3 spec encodes the bytes type to be a base64 encoded string
when marshalling to jsonpb. This adds support to jsonpb to correctly
encode go types in the format `type Bytes []byte` into base64 instead of
being represented as an uint8 array.

I also submitted the same changes to golang/protobuf but since they're being slow I'm also submitting it here.

https://github.com/golang/protobuf/pull/188


awalterschulze:
I would prefer to wait for them.
Sorry but as a fork I prefer to try and keep things simple.
If they reject your request, you are welcome to make the pull request here again.

awalterschulze:
I think waiting for me to merge all the way up to https://github.com/golang/protobuf/commit/3b06fc7a4cad73efce5fe6217ab6c33e7231ab4a
is going to take a while, since I am waiting for proper support for WTKs before merging further.

I think the easiest would be to adjust this pull request to also incorporate the change in proto/text.go

And then if you are ok with giving over your copyright.  I'll merge it.

Oh and keep the test :)

d4l3k:
I just changed to the tests to pass on go versions other than 1.6. 

The change in proto/text.go is actually already in gogo so that shouldn't be a problem.

I'm fine with giving over my copyright.

awalterschulze:
I think go1.3 is less important, but go1.4 is still used, since it was the last c implementation of the runtime and it is used to compile the new go versions.
I would like to continue to support go1.4
So if you can fix it I will merge this.

tamird:
There's a failing test here. Also, the tests being added should use code generated by gogoproto instead of the manually generated code currently included.

jsonpb/jsonpb_test.go
```diff
@@ -264,6 +264,18 @@ var (
 		`}`
 )
 
+// Bytes is used to test that []byte type aliases are serialized to base64.
+type Bytes []byte
+
+// BytesMessage is used to test that []byte type aliases are serialized to base64.
+type BytesMessage struct {
```
awalterschulze: Aha yes it would be better if this message was generated by gogoprotobuf.  Thank you @tamird

d4l3k:
I've updated the code to generate the protobuf used in tests as well as made the unmarshalling code compatible with pre-go1.5.

For some reason (which I don't see listed in the [go1.5 release notes](https://golang.org/doc/go1.5)) this example code only works with >1.5 versions.
```go
package main

import (
	"encoding/json"
	"log"
)

type Bytes []byte

func main() {
	str := Bytes("foo")
	body, err := json.Marshal(str)
	if err != nil {
		log.Fatal(err)
	}
	var out Bytes
	if err := json.Unmarshal(body, &out); err != nil {
		log.Fatal(err)
	}
	log.Printf("%s %s %s", str, body, out)
}
```
On earlier versions you get
```
2016/06/07 11:07:22 json: cannot unmarshal string into Go value of type main.Bytes
exit status 1
```

tamird:
Here's the Go 1.5 change, if you want to include it in the comment. https://github.com/golang/go/commit/4302fd0409da5e4f1d71471a6770dacdc3301197

tamird:
https://github.com/golang/go/commit/c60707b14d6be26bf4213114d13070bff00d0b0a is another change that was subsequently made. Seems to me that the code in this PR has the same problem?

jsonpb/jsonpb_test_proto/bytes.go
```diff
@@ -0,0 +1,4 @@
+package jsonpb
+
+// Bytes is used to test that []byte type aliases are serialized to base64.
+type Bytes []byte
```
tamird: can we use an alias of `byte` here?

d4l3k:
Added it to the comment, added `type Byte byte` and switched to using SetBytes

tamird:
LGTM

@awalterschulze can we haz a merge?

jsonpb/jsonpb_test_proto/test_objects.proto
```diff
@@ -33,6 +33,8 @@ syntax = "proto2";
 
 package jsonpb;
 
+import weak "gogoproto/gogo.proto";
```
awalterschulze: What is import weak?
Is there a way we could use this to not have the proto_path be so complicated i.e. not have the descriptor.proto and gogo.proto be in the proto_path, but still be able to use the gogoproto extensions?
Because that would be amazing.

awalterschulze:
I haz had a look zee and I haz mergzed.

PR #187, opened 2016-06-16
Bump the generatedCodeVersion to 3 so generated code works with grpc.SupportPackageIsVersion3.

beejeebus:
Generated code was failing to build with the latest code from https://github.com/grpc/grpc-go


awalterschulze:
What about the rest of the commit this is actually part of, please see the issue.

PR #190, opened 2016-07-24
io: allow to skip messages

dennwc:
Added API to io package to skip messages without decoding.

### DIFF
--- a/io/io.go
+++ b/io/io.go
@@ -51,6 +51,16 @@ type ReadCloser interface {
 	io.Closer
 }
 
+type Skiper interface {
+	SkipMsg() error
+}
+
+type ReadSkipCloser interface {
+	Reader
+	Skiper
+	io.Closer
+}
+
 type marshaler interface {
 	MarshalTo(data []byte) (n int, err error)
 }
--- a/io/io_test.go
+++ b/io/io_test.go
@@ -39,8 +39,7 @@ import (
 	"time"
 )
 
-func iotest(writer io.WriteCloser, reader io.ReadCloser) error {
-	size := 1000
+func writetest(writer io.WriteCloser, size int) ([]*test.NinOptNative, error) {
 	msgs := make([]*test.NinOptNative, size)
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
 	for i := range msgs {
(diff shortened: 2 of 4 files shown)


awalterschulze:
Can you maybe tell me about the usecase?

dennwc:
It mostly about reading large (tens of GBs) protobuf files from disk in parallel. Few routines are skipping a number of messages to start from unique offset. Also the main routine uses this to count a total number of messages before processing.

But one may use it in other ways: skip malformed/unknown messages, skip large messages. But for the last use case it needs to handle the case when size was already read, and ErrShortBuffer was returned. I can add this functionality to cover more use cases, if you like.

awalterschulze:
I am not convinced that this is such a popular use case, combined with the fact that this is very little code that is well contained, I think you can copy the library and create your own library with the modified code.

Also I think if you have a file that implements https://golang.org/pkg/io/#Seeker you can more efficiently skip messages.

Sorry I hope this answer is ok, but if you would like to discuss more, please go ahead.

dennwc:
@awalterschulze I agree that copying library is ok. But I think since library uses bufio internally it should expose it in some way, because user looses control of the reader passed in.

awalterschulze:
Interesting, but I didn't really want to expose the implementation detail to the user.

dennwc:
@awalterschulze That's exactly what I've assumed, and that's why I added method for skipping messages. Using the package will make Reads unpredictable (underlying reader will not stop at the message boundaries), and there is no way to fix this right now, except of copying the sources.

And copying is sort of bad also, because the lib is all about performance, thus I assume things may change under the hood. I'll need to maintain an updated copy, then.

Regarding Seek, in this change I assumed the bad case - the stream is from the network or it's compressed, thus no chance to use it. But I can add the support for it.

awalterschulze:
No in my case the stream can be from the network, so I can't change it to incorporate seek, but your copy can.
This library has not changed much in its lifetime.  I think this is quite safe to copy.

dennwc:
Sorry, I meant that I can't use seek either because of the compression.

In any case, the decision was made, so I'll just copy the library.

awalterschulze:
Aha ok yes compression can make it quite hard, I didn't realise it was compressed :(

Yes I am pretty sure it won't be a lot of work to maintain.  Maybe you can look at updating it once every two years.  Its a pretty stable library.

PR #203, opened 2016-09-24
CamelCase methodName consistently

### LINKED ISSUE #202: Inconsistent method naming for snake_case method names
```
syntax = "proto3";

package db_dealer;

message db_dealer__utolso_mod__input {
}

message db_dealer__utolso_mod__output {
        string ret = 1;
}

service db_dealer {
        rpc db_dealer__utolso_mod (db_dealer__utolso_mod__input) returns (db_dealer__utolso_mod__output) {}
}
```

Result of `protoc --gofast_out=plugins=grpc:. proto.proto`:

```
func (c *dbDealerClient) DbDealer_UtolsoMod(ctx context.Context, in *DbDealer_UtolsoMod_Input, opts ...grpc.CallOption) (*DbDealer_UtolsoMod_Output, error) {
        out := new(DbDealer_UtolsoMod_Output)
        err := grpc.Invoke(ctx, "/db_dealer.db_dealer/db_dealer__utolso_mod", in, out, c.cc, opts...)
        if err != nil {
                return nil, err
        }
        return out, nil
}

var _DbDealer_serviceDesc = grpc.ServiceDesc{
        ServiceName: "db_dealer.db_dealer",
        HandlerType: (*DbDealerServer)(nil),
        Methods: []grpc.MethodDesc{
                {
                        MethodName: "db_dealer__utolso_mod",
                        Handler:    _DbDealer_DbDealer_UtolsoMod_Handler,
                },
        },
        Streams:  []grpc.StreamDesc{},
        Metadata: fileDescriptorDbDealer,
}
```

The problem is that https://github.com/gogo/protobuf/blob/master/protoc-gen-gogo/grpc/grpc.go#L281 misses generator.CamelCase, but https://github.com/gogo/protobuf/blob/master/protoc-gen-gogo/grpc/grpc.go#L383 has.

This is not a problem iff the .proto file follows the Style Guide which advises the use of CamelCase service- and method names.

Found in https://github.com/gogo/letmegrpc/issues/25

tgulacsi:
Fixes #202 and https://github.com/gogo/letmegrpc/issues/25

### DIFF
--- a/protoc-gen-gogo/grpc/grpc.go
+++ b/protoc-gen-gogo/grpc/grpc.go
@@ -232,7 +232,7 @@ func (g *grpc) generateService(file *generator.FileDescriptor, service *pb.Servi
 			continue
 		}
 		g.P("{")
-		g.P("MethodName: ", strconv.Quote(method.GetName()), ",")
+		g.P("MethodName: ", strconv.Quote(generator.CamelCase(method.GetName())), ",")
 		g.P("Handler: ", handlerNames[i], ",")
 		g.P("},")
 	}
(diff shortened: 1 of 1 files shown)


awalterschulze:
Would you mind modifying something in ./test/grpc.proto to not be a camelcase name so that the current code breaks and that it is fixed with this change?

awalterschulze:
Ok no sorry, this code is forked from golang/protobuf.

This fix should happen in letmegrpc and not in gogoprotobuf.

PR #214, opened 2016-10-14
Fixes panic message in enumstringer plugin

atombender:
If you enable `gogoproto.enum_stringer_all`, `protoc-gen-gofast` fail with:

```
panic: old enum string method needs to be disabled, please use gogoproto.old_enum_stringer or gogoproto.old_enum_string_all and set it to false
```

But those options don't exist. I've fixed the panic message with the right name, and made it more readable.


awalterschulze:
You can add yourself to the authors and/or contributors as appropriate if you want.

atombender:
Not necessary.

awalterschulze:
Thank you very much :)

Issue #174, opened 2016-05-11
Adding a new UnsafeMarshallerWithArenaAllocation


smarterclayton:
I've been experimenting with an unsafe arena allocation technique for generated gogoprotobuf unmarshallers that avoids `[]byte -> string` allocations during decode, by leveraging the technique Vitess describes here: https://github.com/youtube/vitess/blob/master/go/hack/hack.go

It has the potential to dramatically improve the speed of decodes for any object with more than a few strings, but requires a slightly more complex Unmarshal signature to pull off.  Would you accept a ReallyUnsafeMarshaller that allowed the unmarshal caller to pass an Arena object in to the unmarshaller?  Because of the signature change it would be different than the normal unmarshaller, and callers would need to really understand that Go might change the semantics around strings in an incompatible way in the future.  On the other hand... performance...

E.g.

```
f.Unmarshal(data, &proto.Options{Arena: proto.NewArena(len(data) / 2)})

func (f *foo) Unmarshal(data []byte, o *proto.Options) {
  ...
  switch field
  case 1:
     ...
     m.AString = o.Arena.NewString(data[i:l])
     ...
  }
}
```

Picking the arena size relative to the data would be important - guessing based on your object what the percentage of data is part of strings.

awalterschulze:
Have you seen this issue
https://github.com/gogo/protobuf/issues/54

What do you think about the effects of a moving garbage collector?

smarterclayton:
I saw the issue - I *think* the moving garbage collector would still be safe here because we still have a concrete references first to the overall string, and every subsequent string is part of that original object.  However, I am not an expert on the implications so if we think it would be totally broken in the presence of a moving GC I'll give it up (although I haven't seen movement on that).

awalterschulze:
I am not an expert either, I am just a little cautious.
What would be the difference in simply unmarshaling into a byteslice and a proto.Arena?

awalterschulze:
Looks like not much its just adds a little abstraction.

awalterschulze:
Is there some way we could benefit more types than just strings?
It would be nice if we could do all the pointer management ourselves and have even less of a garbage collector impact.
That is what I was trying to do in #54 
But when I push it there I get really scared about the moving garbage collector.
Maybe we can ask someone who knows?

smarterclayton:
[...]

PR #217, opened 2016-10-19
Experiment: Add an unsafe "borrow" unmarshaler that avoids some allocations

smarterclayton:
This PR adds a new unsafe_borrow_unmarshaler tag that is more efficient on string or byte heavy objects during unmarshal, with the caveat that the caller must copy the buffer prior to passing it to Unmarshal() and that total memory allocated may go up.

The borrow unmarshaler avoids the []byte -> string cast Go performs to ensure strings are immutable via the unsafe package, relying on the caller to have provided an immutable buffer. For objects with 30-50% of fields being strings, this offers a substantial CPU improvement above 20% and a corresponding reduction in allocations.

The downside is that simple objects will not perform as well, and due to the signature of .Unmarshal() it is difficult to hide the "copy" operation on a per type basis (although not impossible, and could be done by splitting `UnmarshalBorrow()` and `Unmarshal()` for all types).

Added a new string heavy type and showed the improvement.  Many of the existing number heavy types were unchanged - so this is not a clear win for all types by far.

```
$ benchcmp /tmp/old /tmp/new
benchmark                                old ns/op     new ns/op     delta
BenchmarkStringHeavyProtoUnmarshal-8     945           728           -22.96%

benchmark                                old MB/s     new MB/s     speedup
BenchmarkStringHeavyProtoUnmarshal-8     412.41       535.12       1.30x

benchmark                                old allocs     new allocs     delta
BenchmarkStringHeavyProtoUnmarshal-8     14             8              -42.86%

benchmark                                old bytes     new bytes     delta
BenchmarkStringHeavyProtoUnmarshal-8     648           677           +4.48%
```

If this seems like a reasonable addition I can round out the remaining types to handle and add any necessary testing and doc.


smarterclayton:
Implements / experiments with #174, but in a more useful direction (arena is much more complicated for arguably less benefit).

awalterschulze:
First off I like this, but lets just discuss this a bit again.  Maybe there is a bigger speedup to be gained.
It seems what you are trying to achieve here is striving towards a zero copy unmarshaler.

Maybe you can give a use case where the user will always have copied the buffer before doing the Unmarshal?

You could probably also not copy any fixed64, fixed32, double or floats using unsafe?

How would you propose this split of Unmarshal and UnmarshalBorrow?

smarterclayton:
> Maybe you can give a use case where the user will always have copied the buffer before doing the Unmarshal?

The user has code that wraps the use of unmarshal, usually.  I think the caller, to leverage zero-copy, has to know that he is using zero copy.  Ultimately any unmarshal tree that has any "zero-copy" unmarshaler has to be treated as completed zero-copy.

It's possible to avoid copying a fair number of values by direct byte manipulation.  However, I'm not sure that's actually faster in all cases since there is no allocation present.

For a split, I could imagine this marshaler providing the following signatures:

```
function (t *TYPE) Unmarshal(data []byte) error {
  copied := make([]byte, len(data))
  copy(copied, data)
  return t.UnmarshalBorrow(copied)
}

func (t *TYPE) UnmarshalBorrow(data []byte) error {
  ... normal generated borrower
}
```

Naive clients still call unmarshal and take the hit (but you can selective enable the borrow unmarshaler only for types that matter).  Sophisticated clients either call top level UnmarshalBorrow (because they have an immutable data object already) or just Unmarshal.  The generator would need to try to use the borrow unmarshaler if present all the way down.  However, if you had:

```
type 1 - supports borrow
  type 2 - doesn't support borrow
    type 3 - supports borrow
```

You'd end up doing two more copies than you need and end up losing the benefit.  So a warning / easy way to know that you're mixing borrow and unborrowed generators.

awalterschulze:
Ok doesn't seem like my reply is getting through.
So here is a possible duplicate reply.

```
The user has code that wraps the use of unmarshal, usually. I think the caller, to leverage zero-copy, has to know that he is using zero copy. Ultimately any unmarshal tree that has any "zero-copy" unmarshaler has to be treated as completely zero-copy.
```

I am still trying to imagine some example where someone has a buffer that they are not reusing, because only then they will be able to use the borrow unmarshaler.  I am sure such a scenario exists.  I just want to know of one. 

```
It's possible to avoid copying a fair number of values by direct byte manipulation. However, I'm not sure that's actually faster in all cases since there is no allocation present.
```

Fair enough. Good point.

```
For a split, I could imagine this marshaler providing the following signatures:
```

I think this will slow down the original Unmarshal method, don't you? 

```
You'd end up doing two more copies than you need and end up losing the benefit. So a warning / easy way to know that you're mixing borrow and unborrowed generators.
```

Ok so lets rather have the UnmarshalBorrow method assume that its children also have an UnmarshalBorrow method.  Or are we going to type switch?  Not copying is probably a bigger time safer than a type assert is a time waster.

smarterclayton:
GitHub is eating email responses over the last week or so - happening to me too.

> I am still trying to imagine some example where someone has a buffer that they are not reusing, because only then they will be able to use the borrow unmarshaler.

It would almost always be explicit (someone says copy) or naive (just using bad code).  So zero-copy tends to be something a user is conscious of.  However, the caller of `.Unmarshal*()` can sometimes know that the buffer is safe to use without copy (because due to higher level design in the program the buffer is unique, such as from `ioutil.ReadAll()` or from a frame decoding stream).  So there are cases where a copy is unnecessary because the use case always ends up with you having a uniquely referenced buffer.

> Or are we going to type switch? Not copying is probably a bigger time safer than a type assert is a time waster.

At the highest level, this could be looked at like a compiler optimization - given what we know about a type, is the borrow unmarshal worth it?  Some heuristics:

* Does the message contain enough strings/bytes relative to other fields to justify the cost of a copy (cost of copy is 1 allocation + CPU proportional to message size)
* Is the message included in another message that meets the above criteria?

The other tradeoff is code size - we could be time efficient by generating both paths (one that generates Unmarshal, and one that generates UnmarshalBorrow) but that would slow compilation and increase code size.  

I think in general that the message type alone determines whether borrow is worth it just based on the initial expectation.  I.e.:

```
// never worth generating borrow
message Foo {
  optional int32 field1 = 1
  optional int32 field2 = 2
}

// never worth generating borrow
message Bar {
  optional string field1 = 1
}

// probably always worth generating borrow
message Baz {
  repeated string field1 = 1
}

// probably always worth generating borrow
message Biz {
  repeated Bar field1 = 1
}
```

awalterschulze:
> It would almost always be explicit (someone says copy) or naive (just using bad code). So zero-copy tends to be something a user is conscious of. However, the caller of .Unmarshal*() can sometimes know that the buffer is safe to use without copy (because due to higher level design in the program the buffer is unique, such as from ioutil.ReadAll() or from a frame decoding stream). So there are cases where a copy is unnecessary because the use case always ends up with you having a uniquely referenced buffer.

Ok cool so when using ioutil.ReadAll like in an http request.
Also maybe another is when you are reusing the input buffer, but once you have unmarshaled you process the message fully before unmarshaling the next message.

> At the highest level, this could be looked at like a compiler optimization - given what we know about a type, is the borrow unmarshal worth it? Some heuristics:

I think these heuristics should be left up to the user, since the user already has to make an education decision about when to use UnmarshalBorrow.  I don't think we can decide when to generate this for him. 

The fact that its up to the user's logic to decide when to use UnmarshalBorrow means than Unmarshal can never call into UnmarshalBorrow.

UnmarshalBorrow is never slower than Unmarshal, but in the case where the struct has no strings or bytes, the code is almost identical and increases the code size for seemingly no good reason.

But since Unmarshal can never call into UnmarshalBorrow, we need UnmarshalBorrow calls all the way down once the user has started calling UnmarshalBorrow.

An alternative is to have an `UnmarshalCopy(buf []byte, copy bool) error` method which can then be reused by both Unmarshal and UnmarshalBorrow.  This way we can UnmarshalBorrow all the way down and only when appropriate and code size is kept to a minimum when both Unmarshal and UnmarshalBorrow is generated.  Unfortunately when only one is generated there will be a if clause (for strings and bytes) which will never be executed, but this is probably an ok trade off.

customtype is a problem though, since it only has an `Unmarshal(buf []byte) error` method, but I guess we leave this and don't optimize this as well.

awalterschulze:
On 20 October 2016 at 06:05, Clayton Coleman <notifications@github.com>
wrote:

> Maybe you can give a use case where the user will always have copied the
> buffer before doing the Unmarshal?
>
> The user has code that wraps the use of unmarshal, usually. I think the
> caller, to leverage zero-copy, has to know that he is using zero copy.
> Ultimately any unmarshal tree that has any "zero-copy" unmarshaler has to
> be treated as completed zero-copy.
>
I am still trying to imagine some example where someone has a buffer that
they are not reusing, because only then they will be able to use the borrow
unmarshaler.  I am sure such a scenario exists.  I just want to know of
one.

> It's possible to avoid copying a fair number of values by direct byte
> manipulation. However, I'm not sure that's actually faster in all cases
> since there is no allocation present.
>
Fair enough. Good point.

> For a split, I could imagine this marshaler providing the following
> signatures:
>
> function (t *TYPE) Unmarshal(data []byte) error {
>   copied := make([]byte, len(data))
>   copy(copied, data)
>   return t.UnmarshalBorrow(copied)
> }
>
> func (t *TYPE) UnmarshalBorrow(data []byte) error {
>   ... normal generated borrower
> }
>
> I think this will slow down the original Unmarshal method, don't you?

>  Naive clients still call unmarshal and take the hit (but you can
> selective enable the borrow unmarshaler only for types that matter).
> Sophisticated clients either call top level UnmarshalBorrow (because they
> have an immutable data object already) or just Unmarshal. The generator
> would need to try to use the borrow unmarshaler if present all the way
> down. However, if you had:
>
> type 1 - supports borrow
>   type 2 - doesn't support borrow
>     type 3 - supports borrow
>
> You'd end up doing two more copies than you need and end up losing the
> benefit. So a warning / easy way to know that you're mixing borrow and
> unborrowed generators.
>
Ok so lets rather have the UnmarshalBorrow method assume that its children
also have an UnmarshalBorrow method.  Or are we going to type switch?  Not
copying is probably a bigger time safer than a type assert is a time waster.

awalterschulze:
I am closing this because of inactivity, but just comment if you would like it reopened.

pjvds:
@smarterclayton There are multiple issues registered where people would love to see zero copy unmarshal support; are you still interested in getting this one ready for a pull?

PR #223, opened 2016-10-25
Add Compare to oneof Interfaces if needed

### LINKED ISSUE #221: Using compare_all with proto oneof generates invalid Interface
Hi,

When using a message which contains a message with a `oneof` variable and using the setting to generate compare methods, the generated interface for the variable does not have the Compare() function generated as part of the interface.

Here is a gist which reproduces the error:

https://gist.github.com/johanbrandhorst/783e0908940a2f81d7aef7b1ac93472e

The offending line is my.pb.go:44:

```go
type isValue_Type interface {
	isValue_Type()
}
```

Should be

```go
type isValue_Type interface {
	isValue_Type()
        Compare(interface{}) int
}
```

johanbrandhorst:
I think this should fix #221.


awalterschulze:
Maybe you can get away with simply adding the compare extension to test/oneof/one.proto for a test.

Could you also reclone or something, since I have squashed all your previous commits.

johanbrandhorst:
Cool, I'm pretty new to open source contributions, I'll definitely try and sort out the commit log.

johanbrandhorst:
I'll make a new pull request

Issue #199, opened 2016-09-16
Known Issue: proto.MarshalTextString cannot handle custom types with Bytes method


EricLagergren:
If a custom type implements `proto.raw` (i.e., has a `Bytes() []byte` method) `proto.TextMarshaler.writeStruct` will call it, causing invalid bytes to be written.

For example, I have a UUID type with a `Bytes() []byte` method which returns the UUID's text format as a slice of bytes (it reduces allocations for other methods like `MarshalText`).

I use this as a custom type and when I do I get the errors [seen here](https://github.com/gogo/protobuf/issues/198#issuecomment-247465177).

I'm not sure what your take is on this. Either there should be some docs (I know you said before custom type has issues, so if I could help let me know) or instead of `Bytes() []byte` the interface should be `type raw interface { bytes() []byte }` so no external types can implement it.

EricLagergren:
Oh. And this is tangentially related: if a custom type implements `encoding.TextMarshaler` `proto.TextMarshaler.Text` does not check the error return of the call to `Marshal` meaning if the custom type's `MarshalText` method returns an error the test will fail.

```
session: <
  id: "Kp3ro_dqJvf4FHTcqTHbsjLMCl8NYkZdWpfu8jqKLTWt2Z2CBqwoD3poXhYYaXK-CgI9tSUbGndFpFS6S0bRlg=="
  token_hash: "e\260\226N)\343\242\322\262Q\263\002\227\005\245\244\264l\035\025\014c\314!\"\010?7\265\253\201\372b\327\274>>\235m\303\rA\203\302\236\006\245"
  principal: <
    id: "\366\211I\201\327x\221 ._w\273l\276\177\020"
    name: "jQsrF340cm7mPOcG1uRvsSeP1SU4aLCvoE5XcIdjgFGi50gUaUMS0uiHqR3ak2OU"
    type: sermoadmin
    domain: "hI4QYd5i1FMJlbFl"
    template: "B0pgKupYaI98wljnt9DoDkgHTlJR6FZnL3ypNEgpKcd2Z4Frp0uointY9hX17i"
  >
  email: "ZsvtMVQ"
  user_id: "\231\350l$0:U\315Y\022\025\r5\276pr"
  first_name: "60pN7eVYbkykbKISujGrCqrOE224vlyFX1EhQU2rA7WTUq96OsQolqR69KdYvSbWCMmL5oKNwT9"
  start: <
    seconds: 5253274347
    nanos: 988209534
  >
[...]

PR #227, opened 2016-11-11
text: allow customtype to have a Bytes() method

tamird:
If encoding.TextMarshaler is implemented, it will be preferred.

Related to #199. @awalterschulze cc @bdarnell

### DIFF
--- a/extensions.md
+++ b/extensions.md
@@ -50,9 +50,9 @@ You might also find that basic structs that started their life as part of an API
 <tr><td><a href="https://github.com/gogo/protobuf/blob/master/test/types/types.proto">stdduration</a></td><td> Duration Field </td><td> bool </td><td>Changes the Well Known Duration Type to time.Duration</td><td>Duration</td></tr>
 </table>
 
-`Warning about nullable: according to the Protocol Buffer specification, you should be able to tell whether a field is set or unset. With the option nullable=false this feature is lost, since your non-nullable fields will always be set.` 
+`Warning about nullable: according to the Protocol Buffer specification, you should be able to tell whether a field is set or unset. With the option nullable=false this feature is lost, since your non-nullable fields will always be set.`
 
-# Goprotobuf Compatibility 
+# Goprotobuf Compatibility
 
 Gogoprotobuf is compatible with Goprotobuf, because it is compatible with protocol buffers (see the section on tests below).
 
@@ -118,7 +118,7 @@ Other serialization formats like xml and json typically use reflect to marshal a
 
 <a href="https://groups.google.com/forum/#!topic/gogoprotobuf/xmFnqAS6MIc">Here is a longer explanation of jsontag and moretags</a>
 
-# File Options 
+# File Options
 
 Each of the boolean message and enum extensions also have a file extension:
 
@@ -155,7 +155,7 @@ Each of these are the same as their Message Option counterparts, except they app
 
 # Tests
 
-  * The normal barrage of tests are run with: `make tests` 
+  * The normal barrage of tests are run with: `make tests`
   * A few weird tests: `make testall`
   * Tests for compatibility with [golang/protobuf](https://github.com/golang/protobuf) are handled by a different project [harmonytests](https://github.com/gogo/harmonytests), since it requires goprotobuf.
   * Cross version tests are made with [Travis CI](https://travis-ci.org/gogo/protobuf).
--- a/test/combos/both/uuid.go
+++ b/test/combos/both/uuid.go
@@ -47,6 +47,10 @@ func PutLittleEndianUint64(b []byte, offset int, v uint64) {
 
 type Uuid []byte
 
+func (uuid Uuid) Bytes() []byte {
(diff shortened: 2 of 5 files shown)


awalterschulze:
I think the fact that time.Time also implements encoding.TextMarshaler is causing the tests to fail.

Also how about a test which includes a customtype with a Bytes method.

You can also add yourself or your company to the AUTHORS file if you want some copyright.

tamird:
ugh, that's pretty annoying. any ideas on how to fix this?

awalterschulze:
How about simply checking whether the type is time.Time or time.Duration and reverting to the default behaviour for those?

You can use props.StdTime and props.StdDuration to check.

tamird:
@awalterschulze implemented your suggestion, though I don't feel good about it :(

awalterschulze:
Do you have a better plan?
It doesn't look that bad to me.

tamird:
Nah, this is probably fine. Merge at will!

tamird:
Er, hang on, I think we still need a test.

awalterschulze:
Yes please :)

tamird:
Wrote a test! Unfortunately, `raw` expects a serialized proto. Opened https://github.com/golang/protobuf/issues/311 to discuss removal, but I think we can just go ahead and remove it from gogo/protobuf. WDYT?

tamird:
Pushed an update; let me know if that looks good to you and I'll squash it down.

awalterschulze:
The build is failing because you haven't merged in a long time.
Go1.6 failures are permitted.

awalterschulze:
Maybe I am missing something, but this diff is look really weird.
It seems like you have only deleted some code concerning raw and added a bytes method to some Uuid types.  Maybe my memory is wonky, but I remember this pull request going a totally different way with extra checking for stdtime and stdduration.

awalterschulze:
Lets first see what happens with golang/protobuf before simply removing it.

tamird:
> It seems like you have only deleted some code concerning raw and added a bytes method to some Uuid types. Maybe my memory is wonky, but I remember this pull request going a totally different way with extra checking for stdtime and stdduration.

Right; the old approach was to whitelist `stdtime` and `stdduration` so that they don't fall into the textmarshaler case, but that's no longer necessary now that `raw` is simply removed. I added `Bytes() []byte` to `Uuid` to demonstrate that the `Bytes() []byte` method is no longer treated specially.

awalterschulze:
Yes but then I loose a feature that golang/protobuf has.
I have not done this before.
Not without the user explicitly asking for it using an extension of vanity binary.

tamird:
What's the feature? The ability to pass "raw" messages? It's not documented, and according to @bcmills exists only to support an internal Google use case.

awalterschulze:
Being used by google internally is not not a goal of gogoprotobuf.

tamird:
I know! That's what I'm saying; `raw` is only used by Google internally.

awalterschulze:
DId you spot my double negative?

tamird:
Heh, I guess I did not. Well, I guess that's awkward, then.

tamird:
Note that https://github.com/golang/protobuf/issues/311 was "fixed" in https://github.com/golang/protobuf/pull/482, which will be merged to master eventually (https://github.com/golang/protobuf/issues/479). Once that happens, and the changes are pulled into gogo/protobuf, I'll rebase this.

awalterschulze:
I will also create a dev branch, where I will start merging those changes into.  Hopefully that will also help.

tamird:
Ok. Let me know when that branch is ready and I'll fix this up.

On Feb 2, 2018 11:53, "Walter Schulze" <notifications@github.com> wrote:

I will also create a dev branch, where I will start merging those changes
into. Hopefully that will also help.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/gogo/protobuf/pull/227#issuecomment-362636372>, or mute
the thread

awalterschulze:
The dev branch is finally ready :)

tamird:
OK, rebased. I think there's really nothing to do here, now, since upstream removed their overly broad type. All that's left here is a test.

awalterschulze:
Is the writeRaw function now unused code? Maybe the removal of this
function should be upstreamed?

awalterschulze:
I think there are already tests for Uuid and timestamp that do text marshaling:

  - TestNidOptCustomProtoText
  - TestStdTypesProtoText

So by you adding a Bytes methods to Uuid, I think it is getting tested.

awalterschulze:
So if you can upstream writeRaw instead of removing it here and add a comment to the Bytes method for why it is there, just so that we don't remove it, thinking it is something trivial.  Then this is ready to merge, no test needed :)

awalterschulze:
I have pulled it in :)

tamird:
Updated.

awalterschulze:
Can't believe this is finally fixed.
Thank you for your perseverance

PR #229, opened 2016-11-23
file option for prefix to proto.Register{Type,Enum}

IncSW:
Added file option `gogoproto.proto_register_prefix` for `proto.Register{Type,Enum}` name prefix.

We use services where every has own pb, and sometimes messages and/or enums has similar name (and similar package name).


awalterschulze:
enums and messages are registered using different methods and with their respective packagenames.  You must mean that you have multiple packages with the same name that also has messages with the same name.  

Have you first tried explaining your problem to golang/protobuf?

My opinion would be to simply change you package names, but maybe I am not understanding the problem?

IncSW:
Yes, we can solve it with changing package names or changing message names.
But it mean what services must know about each other too much (and developers too, when developers has own service).

When services with are same (standard for project) structure they have same packege names for api. And some messages may have similar name but it not mistake and may be correct for business logic.

awalterschulze:
Ok I can now see how this can a problem.
Separate projects should be able to choose their message names and package names independently and not cause problems for each other proto encodings.

Lets first explore the option that fixes this for more people.
Have you logged an issue against golang/protobuf?

Huskie-git:
panic: proto: duplicate enum registered: pubsub.pb.TopicDescriptor_AuthOpts_AuthMode

goroutine 1 [running]:
github.com/gogo/protobuf/proto.RegisterEnum(...)
        /Users/macos/go/pkg/mod/github.com/gogo/protobuf@v1.3.2/proto/properties.go:520
github.com/libp2p/go-libp2p-pubsub/pb.init.0()
        /Users/macos/go/pkg/mod/github.com/libp2p/go-libp2p-pubsub@v0.4.1/pb/rpc.pb.go:830 +0x6a5

Huskie-git:
two package libp2p@13

PR #233, opened 2016-12-01
less go versions are tested because travis breaks for go 1.5.4 and go…

awalterschulze:
… 1.6.1, even on old commits and reverted commits which worked before

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,10 @@ language: go
 
 go:
   - 1.7.1
+  - 1.6.3
+  - 1.5.4
+
+matrix:
+  allow_failures:
+    - 1.6.3
+    - 1.5.4
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,4 @@ script:
 language: go
 
 go:
-  - 1.5.4
-  - 1.6.3
   - 1.7.1
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,11 @@ script:
 language: go
 
 go:
-  - 1.7.1
-  - 1.6.3
   - 1.5.4
+  - 1.6.3
+  - 1.7.1
 
 matrix:
   allow_failures:
-    - 1.6.3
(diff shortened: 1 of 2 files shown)


tamird:
You should make these allowed failures rather than removing them entirely.

awalterschulze:
Is there a way to do that in travis?

awalterschulze:
Lets see what happens 
http://stackoverflow.com/questions/40919256/older-go-versions-on-travis-are-killed-in-the-middle-of-running-tests

tamird:
https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail

awalterschulze:
We can bring these tests back in a pull request, once they pass

awalterschulze:
Hmmm, ok but we aren't using matrix
My travis knowledge is pretty weak

awalterschulze:
Ok I think I figured it out.  Lets see what happens

.travis.yml
```diff
@@ -15,6 +15,11 @@ script:
 language: go
 
 go:
-  - 1.5.4
-  - 1.6.3
```
tamird: you should probably restore this bit and squash your commits before merging.

awalterschulze: fair :)

awalterschulze:
Seems like its still failing, even though the fails are allowed.

.travis.yml
```diff
@@ -15,6 +15,11 @@ script:
...
   - 1.7.1
+  - 1.6.3
+  - 1.5.4
+
+matrix:
+  allow_failures:
+    - 1.6.3
```
tamird: these need to be `- go: 1.6.3`

awalterschulze: Thanks.

awalterschulze:
Looks like its going to work :)

awalterschulze:
Awesome thanks a lot :)

Issue #201, opened 2016-09-20
customtype of type struct cannot jsonpb unmarshal


ericlagergren:
(Posting b/c you asked for a separate issue.)

See: https://github.com/gogo/protobuf/issues/199#issuecomment-248066262

awalterschulze:
Would you like to attempt a fix?

EricLagergren:
See my comment on the other issue. This one is much simpler so I think it's manageable :)

awalterschulze:
Fixed by https://github.com/gogo/protobuf/pull/239

Issue #132, opened 2016-01-05
jsonpb: No support for unexported fields (and json.Marshaler)


dennwc:
Can be reproduced with custom type:
```Go
package pt

//go:generate protoc --proto_path=$GOPATH/src:. --gogo_out=. types.proto

type Native struct{
	data string
}

// ... omiting methods to satisfy protobuf interfaces.
```
```proto
// types.proto
syntax = "proto3";

package pt;

// Contains custom type definition to bypass generation errors:
// message Native{}
import ".../native.proto";

import "github.com/gogo/protobuf/gogoproto/gogo.proto";

option (gogoproto.sizer_all) = true;
option (gogoproto.marshaler_all) = true;
option (gogoproto.unmarshaler_all) =  true;
option (gogoproto.testgen_all) =  true;
option (gogoproto.populate_all) =  true;
option (gogoproto.equal_all) =  true;

message Item{
    Native sub = 1;
}
```
Auto-generated tests for serialization will fail with panic:
```
panic: reflect.Value.Interface: cannot return value obtained from unexported field or method

reflect.valueInterface(...)
	/usr/local/go/src/reflect/value.go:912 +0xe7
reflect.Value.Interface(...)
	/usr/local/go/src/reflect/value.go:901
github.com/gogo/protobuf/jsonpb.(*Marshaler).marshalValue(...)
	src/github.com/gogo/protobuf/jsonpb/jsonpb.go:371
github.com/gogo/protobuf/jsonpb.(*Marshaler).marshalField(...)
	src/github.com/gogo/protobuf/jsonpb/jsonpb.go:216
github.com/gogo/protobuf/jsonpb.(*Marshaler).marshalObject(...)
	src/github.com/gogo/protobuf/jsonpb/jsonpb.go:147
github.com/gogo/protobuf/jsonpb.(*Marshaler).marshalValue(...)
	src/github.com/gogo/protobuf/jsonpb/jsonpb.go:316
github.com/gogo/protobuf/jsonpb.(*Marshaler).marshalField(...)
	src/github.com/gogo/protobuf/jsonpb/jsonpb.go:216
github.com/gogo/protobuf/jsonpb.(*Marshaler).marshalObject(...)
	src/github.com/gogo/protobuf/jsonpb/jsonpb.go:147
github.com/gogo/protobuf/jsonpb.(*Marshaler).Marshal(...)
	src/github.com/gogo/protobuf/jsonpb/jsonpb.go:75
[...]

Issue #125, opened 2015-12-04
Known Issue: customtype that is another protocol buffer message with gogoprotobuf generated functions and methods is not supported


dennwc:
(Copied from #13)
There is an issue with this combination of options:
```proto
message M {
    repeated bytes arr = 1 [(gogoproto.customtype) = "MyType", (gogoproto.nullable) = false];
}
```
It will generate next code for M:
```Go
type MyType struct{}
func (this *MyType) Equal(that interface{}) bool {
    // ...
   that1, ok := that.(*MyType) // auto-generated function assumes pointer to type 
    if !ok {
        return false
    }
    // ...
}

type M struct {
    Arr []MyType `protobuf:"bytes,1,rep,name=arr,customtype=MyType" json:"arr"`
}

func (this *M) Equal(that interface{}) bool {
    // ...
    if !this.Arr[i].Equal(that1.Arr[i]) { // should be &that1.Arr[i]
        return false
    }
    // ...
}
```
It will fail because `Equal` method will try to cast a parameter to `*MyType`. So, the behavior of `Equal` method is asymmetric - it will try to pass an `Arr[i]` element, but their auto-generated versions assumes, that parameter will be a pointer to the type.

awalterschulze:
This should fix your issue
https://github.com/gogo/protobuf/commit/dcff63233e3f06c3388237a31a95f7b157cf9031

Thank you for the idea for the fix :)

dennwc:
The issue is still there:
- it now assumes non-pointer argument in Equal for custom types for `*.pb.go` files and a pointer argument for `*pb_test.go` files (probably the test plugin was not updated)
- it is still asymmetric: assumes a pointer for auto-generated types and non-pointer for custom types

Can we make Equal to have a concrete parameter type instead of empty interface? It will check everything on compile-time at least)

awalterschulze:
If a type is specified as the parameter an Equaler interface is not possible.
```
type Equaler interface {
  Equal(interface{}) bool
}
```

awalterschulze:
It looks like its working
https://github.com/gogo/protobuf/commit/e492fd34b12d0230755c45aa5fb1e1eea6a84aa9
customtype is supposed to be a type you created yourself and not another protocol buffer.
Obviously it can be, but it you will at least need to edit the Marshal method to not have a pointer receiver.
Also you will need a different NewPopulate method.

awalterschulze:
To fix this will require a backwards incompatible change.

[...]

PR #239, opened 2016-12-09
Allow usage of message types with customtype override

jvshahid:
this pr tries to fix an issue with gogoproto when a customtype is used in conjuction with a message type, e.g.:

``` go
message DesiredLRPSchedulingInfo {                                                                                                                                                              
  ...                                                                                                                                                                                           
  optional ProtoRoutes routes = 1 [(gogoproto.nullable) = false, (gogoproto.customtype) = "Routes"];                                                                                            
  ...                                                                                                                                                                                           
}                                                                                                                                                                                               
                                                                                                                                                                                                
message ProtoRoutes {                                                                                                                                                                           
  map<string, bytes> routes = 1;                                                                                                                                                                
}                                                                                                                                                                                               
```

The idea is to decorate the map in golang but allow non Golang (e.g. Ruby)
clients to be generated with reasonable types. The alternative of using `bytes`
as the type of `routes` in this case generate non Golang marshal methods that
are dumb and won't do any serialization/checking on the values of the `routes`
field.

**note** this pr is just a demonstration and we don't expect it to be merged as
is. we are happy for feedback and will do the necessary updates to get it
merged in.

fyi, the code sample mentioned above is from the [diego bbs models](https://github.com/cloudfoundry/bbs/blob/master/models/desired_lrp.proto#L22) that is used in cloudfoundry


jvshahid:
Looks like the build failed for protoc `v3.1.0`, not sure why. it is hard to understand the output. If someone clarify the reason, we can look into it.

awalterschulze:
Doesn't seem like you regenerated the code.
Its just doing a git diff

jvshahid:
sweet, regenerated the code using protoc 3.1.0 and golang 1.7.1

jvshahid:
@awalterschulze what do you think of the change ?

awalterschulze:
I think we have to keep these issues in mind:
  - https://github.com/gogo/protobuf/issues/132
  - https://github.com/gogo/protobuf/issues/125
  - https://github.com/gogo/protobuf/issues/201

awalterschulze:
Ok 125 does not seem to apply here, since it is trying to have a proto message as the customtype and is still keeping the proto field type as type bytes.

awalterschulze:
Ok maybe none of them are applicable.

awalterschulze:
Basically you just want to be able to have a custom type where the proto field is of some message type and not only of type bytes.
Even though the serialization would look the same, someone from another language could deserialize these bytes into the defined proto message, but the gogo user will be able to use the custom go coded struct.
Am I understanding this correctly?

If so I am pretty sure you'll need more tests.
Would be nice to add a totally separate test struct in thetest.proto
https://github.com/gogo/protobuf/blob/master/test/thetest.proto
This should then generate a combination of tests for you, which use and don't generated code for marshaling and unmarshaling.

awalterschulze:
So yes I am interested in adding this feature :)
But lets see how the tests cope first.

jvshahid:
> Basically you just want to be able to have a custom type where the proto field is of some message type and not only of type bytes.
> Even though the serialization would look the same, someone from another language could deserialize these bytes into the defined proto message, but the gogo user will be able to use the custom go coded struct.
> Am I understanding this correctly?

yes, that's right

> 
> If so I am pretty sure you'll need more tests.
> Would be nice to add a totally separate test struct in thetest.proto
> https://github.com/gogo/protobuf/blob/master/test/thetest.proto
> This should then generate a combination of tests for you, which use and don't generated code for marshaling and unmarshaling.

sweet, i'll work on the tests and get back to you

jvshahid:
so i added the tests as requested.     currently there are two failres. one related to the text representation of
    the proto message, which i think need to be fixed to use the right struct, it is probably doing reflection on the wrong type. the    other is related to the nullability of the go custom type, which i'm not sure
    about.

awalterschulze:
No both are just reflection woes in the jsonpb and prototext libraries respectively.
They are always lots of "fun"

jvshahid:
So I got to fixing the tests for both json and text marshaling. I'm less comfortable with the json marshaling change, since it requires the custom type to implement `MarshalJSON` and will fail with a weird error if the user doesn't provide those. BTW, the same is true for text marshaling, all the weird errors i was getting previously were caused by an invalid `Marshal()` signature, where I had a pointer receiver but the code expected a value receiver. I think there are two examples that demonstrate how to write a custom type but it's not obvious that these functions are required to have the given signature. I think documenting those requirement will be beneficial. Thoughts ?

test/example/example.proto
```diff
@@ -55,7 +55,11 @@ message A {
 message B {
 	option (gogoproto.description) = true;
 	optional A A = 1 [(gogoproto.nullable) = false, (gogoproto.embed) = true];
-	repeated bytes G = 2 [(gogoproto.customtype) = "github.com/gogo/protobuf/test/custom.Uint128", (gogoproto.nullable) = false];
+	optional G G = 2 [(gogoproto.customtype) = "github.com/gogo/protobuf/test/custom.Uint128"];
```
awalterschulze: Lets not change this example or add an extra one

test/thetest.proto
```diff
@@ -624,3 +624,11 @@ message Node {
 	repeated Node Children = 2;
 }
 
+message NonByteCustomType {
+  option (gogoproto.compare) = false;
```
awalterschulze: Is compare not supported?

jsonpb/jsonpb.go
```diff
@@ -494,12 +494,25 @@ func (m *Marshaler) marshalValue(out *errWriter, prop *proto.Properties, v refle
...
+			return nil
+		}
+
 		pm, ok := iface.(proto.Message)
 		if !ok {
 			if prop.CustomType == "" {
 				return fmt.Errorf("%v does not implement proto.Message", v.Type())
 			}
 			t := proto.MessageType(prop.CustomType)
+			// fmt.Printf("################ %T, %T, %s\n", i, t, prop.CustomType)
```
awalterschulze: remove comment

jsonpb/jsonpb.go
```diff
@@ -725,6 +738,15 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 
 	// Handle nested messages.
 	if targetType.Kind() == reflect.Struct {
+		if target.CanAddr() {
+			// fmt.Printf("################ %T\n", target.Interface())
```
awalterschulze: remove comment

awalterschulze:
If you want to add documentation as you mentioned for customtype that will be really really awesome.
I think it might even warrant its own .md file which we can link to from the other docs, if you are up for it?

jvshahid:
Removed the comments, made the message comparable and added a new document for custom types. With hindsight, I should not have squashed the commits to make it easier to review, sorry about that. I'm also not sure if the new documentation fixes #201 ?

test/combos/unmarshaler/thetest.proto
```diff
@@ -624,3 +624,10 @@ message Node {
 	repeated Node Children = 2;
 }
 
+message NonByteCustomType {
+  optional ProtoType Field1 = 1 [(gogoproto.customtype) = "GoType"];
```
awalterschulze: Just thought of this the other day, but can we also have structs which use nullable=false and repeated custom types, similar to NidOptCustom, NinOptCustom, NidRepCustom and NinRepCustom

extensions.md
```diff
@@ -59,6 +59,7 @@ Issues with customtype include:
   * <a href="https://github.com/gogo/protobuf/issues/125">Using a proto message as a customtype is not allowed.</a>
   * <a href="https://github.com/gogo/protobuf/issues/200">cusomtype of type map can not UnmarshalText</a>
   * <a href="https://github.com/gogo/protobuf/issues/201">customtype of type struct cannot jsonpb unmarshal</a>
+  * Custom types are required to implement a handful of methods. See [CustomTypes](custom_types.md) for more information.
```
awalterschulze: I think we can move all these issues to the customtypes.md file and then simply reference the customtypes.md file in the table above.

custom_types.md
```diff
@@ -0,0 +1,54 @@
...
+``` go
+func (r GoType) Marshal() ([]byte, error)
+func (r *GoType) MarshalTo(data []byte) (n int, err error)
+func (r *GoType) Unmarshal(data []byte) error
+
+func (gt GoType) MarshalJSON() ([]byte, error)
+func (gt *GoType) UnmarshalJSON(data []byte) error
+
+func (gt GoType) Compare(other GoType) int // only required if the comparable option is set
+```
```
awalterschulze: What about Equal and NewPopulated, etc.
I think the comment about a method being optional can be added above the method.
You can also get the highlighting to keep on working by simply adding curly braces to the function definitions.
```go
func (r GoType) Marshal() ([]byte, error) {}
```

custom_types.md
```diff
@@ -0,0 +1,54 @@
...
+
+message ProtoType {
+  optional string Field = 1;
+}
+```
+
+or alternatively you can declare the field type in the protocol message to be
+`bytes`:
+
+``` protocol-buffer
```
awalterschulze: I think its simply ```proto

custom_types.md
```diff
@@ -0,0 +1,54 @@
...
+ensure that the custom type marshals and unmarshals to the expected wire
+format. That is, in the first example, gogo protobuf will not attempt to ensure
+that the wire format of `ProtoType` and `GoType` are wire compatible.
+
+## Custom type method signatures
+
+The custom type must define the following methods with the given
+signatures. Assuming the custom type is called `GoType`:
+
+``` go
```
awalterschulze: You can also add a link to your test file `go_type.go`

awalterschulze:
Also you are welcome to add yourself to the AUTHORS or CONTRIBUTORS file as appropriate.

awalterschulze:
Yes I think might fix #201 @ericlagergren what do you think?

awalterschulze:
I really like the docs, I think its really clear and will result in less confusion from users in future.
Thanks a lot for you work on this, I really appreciate it :)

ericlagergren:
Docs look great! I'd suggest using `T` instead of `GoType` since I think it's more idiomatic and well known.

I could give a more in-depth look over on Sunday or Monday, I'm on my phone right now 😉

jvshahid:
Sorry didn't get a chance to look into the comments until now. They all sound reasonable. I'll try to go through them and push the appropriate fixes as soon as I can. Thanks!

jvshahid:
I think I covered all the feedback y'all gave me. I didn't force push to make it easier to review the changes. Once everything is okay I can squash everything and rebase on master. Let me know what you think

custom_types.md
```diff
@@ -20,35 +20,49 @@ message ProtoType {
...
-func (gt GoType) Compare(other GoType) int // only required if the comparable option is set
+// only required if the compare option is set
+func (t T) Compare(other T) int {}
+// only required if the equal option is set
+func (t T) Equal(other T) bool {}
+// only required if populate option is set
+func NewPopulatedT(r randyThetest) *T {}
 ```
+
+Check [t.go](test/t.go) for a full example
```
awalterschulze: Does this link actually work?

awalterschulze:
Really good work.  Thank you very much.
Please resolve the conflicts and submit again.

awalterschulze:
To resolve conflicts I think you simply need to regenerate

jvshahid:
Sweet. Thanks for the thorough review and accepting this pr. BTW, there were minor changes to the proto3 test suite generated files. I'm not sure if that's ok or not. Let me know if they shouldn't have changed and i can look into it.

awalterschulze:
Looks fine.
Thanks again, this is really cool.

Issue #243, opened 2016-12-30
gostring_all generates invalid code if gogoprotobuf is not imported.


ains:
If a file is processed by the gostring plugin (gostring_all in particular, because gostring will then be run on files which don't necessarily have to import gogoprotobuf) the generated code will attempt to use the method GetUnsafeExtensionsMap from "github.com/golang/protobuf/proto" where it doesn't exist, instead of "github.com/gogo/protobuf/proto" where it does.

The issue was introduced in 
https://github.com/gogo/protobuf/commit/6abcf94fd4c97dcb423fdafd42fe9f96ca7e421b#diff-bebe69f1868bace0174c76c29e8d628fR336

Due to the conditional at
https://github.com/gogo/protobuf/blob/master/plugin/gostring/gostring.go#L140

I'll submit a PR which doesn't generate this method if your message does not have extensions or an extension map, which resolves our issue as we don't use this feature, but the underlying issue will remain.

awalterschulze:
Yes extensions are not really supported when gogoprotobuf is not imported, but when extensions are not used in any of your structs gostring should work.
I will appreciate a pull request.

awalterschulze:
Please add yourself to the CONTRIBUTORS or AUTHORS file as appropriate, if you want to.

ains:
Thanks for the quick response!

That PR actually needs more thought than just a quick one liner in the Github web UI, will fix it properly and send one in!

awalterschulze:
Awesome looking forward to it :)

awalterschulze:
fixed by #245

PR #245, opened 2016-12-30
Don't generate unused extensionToGoString method

ains:
Somewhat mitigates #243
But also avoids generating an unused method.

Ideally I think this should either fail fast if `extensionToGoStringUsed && !gogoproto.ImportsGoGoProto(file.FileDescriptorProto)` or import `github.com/gogo/protobuf/proto` implicitly (which might break what people expected after the nodep commit?

### DIFF
--- a/plugin/gostring/gostring.go
+++ b/plugin/gostring/gostring.go
@@ -97,8 +97,10 @@ not print their values, while the generated GoString method will always print al
 package gostring
 
 import (
+	"fmt"
 	"github.com/gogo/protobuf/gogoproto"
 	"github.com/gogo/protobuf/protoc-gen-gogo/generator"
+	"os"
 	"strconv"
 	"strings"
 )
@@ -145,6 +147,7 @@ func (p *gostring) Generate(file *generator.FileDescriptor) {
 	reflectPkg := p.NewImport("reflect")
 	sortKeysPkg := p.NewImport("github.com/gogo/protobuf/sortkeys")
 
+	extensionToGoStringUsed := false
 	for _, message := range file.Messages() {
 		if !p.overwrite && !gogoproto.HasGoString(file.FileDescriptorProto, message.DescriptorProto) {
 			continue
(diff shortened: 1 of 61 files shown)


awalterschulze:
Usually I prefer to print a WARNING, but I think you might be right in this case.  When extensionsToGoString are used and gogoprotobuf is not imported this should just be an error.

ains:
Cool, looks like I've got some tests that need change + I'll add one to check the error case shortly

awalterschulze:
The tests just need to be regenerated.

awalterschulze:
Running `make` in the root of the project should do it.
Given you have protoc 3.1.0 and go 1.7.x

ains:
@awalterschulze mind taking a look?

plugin/gostring/gostring.go
```diff
@@ -338,29 +341,34 @@ func (p *gostring) Generate(file *generator.FileDescriptor) {
...
-	p.Out()
-	p.P(`}`)
-	p.P(`s+=`, stringsPkg.Use(), `.Join(ss, ",") + "})"`)
-	p.P(`return s`)
-	p.Out()
-	p.P(`}`)
-
+	if extensionToGoStringUsed {
+		if !gogoproto.ImportsGoGoProto(file.FileDescriptorProto) {
+			p.Error(fmt.Errorf("Please import gogoproto in '%v' in order to use extensions.", p.localName))
```
awalterschulze: fmt.Fprintf(os.Stderr, "The GoString plugin for messages with extensions requires importing gogoprotobuf.  Please see file %s", file.Name)
os.Exit(1)

awalterschulze:
The rest looks perfect :)
I haven't used p.Error before, but maybe its also fine.

ains:
Switched it to os.Exit and updated the message :)

awalterschulze:
You are welcome to also add yourself to the CONTRIBUTORS or AUTHORS as appropriate?

awalterschulze:
Or I can merge this now?

ains:
That's fine - feel free to merge :)

awalterschulze:
Awesome :)  Thanks for your contribution.

PR #254, opened 2017-01-22
Remove timestamp allocs

bmizerany:
When using gogofaster all of my allocations went away except for stdtime fields. This fixes that. The bench results are in the commit message.


awalterschulze:
Can't we do better than a sync pool here?

awalterschulze:
Basically creating a Timestamp just to Unmarshal it and then convert to another type is a very slow way to simply unmarshal to ints into a time.Time struct.
I just did this to get things working and now that there is ample interest an optimization is warranted.
But we don't need to use any syncing, we can go even faster.

A function can be written to return two ints instead of unmarshaling into an object.
And then these ints can be placed into the time.Time struct.
This should also get rid of the allocs.

You can also turn your benchmark into a test that fails it any allocs occur.

bmizerany:
Noted. I'll look into this tomorrow.

bmizerany:
Okay. Took a quick stab at this tonight and got close to a 10x speed increase after replacing the sync.Pool with a custom function to unmarshal the data. I started to thing making a custom function would/cloud break unmarshaling if anything changed in timestamp.proto since it wouldn't get any updates the generated code would but I concluded that it is unlikely it will change. If anything does break there are enough tests in here to catch it I suppose.

> Full disclosure: I copied much of a code from the generated source in timestamp.go which I was a bit hesitant to do but as I began handwriting it looked very similar so I figured... meh.

types/timestamp_test.go
```diff
@@ -135,3 +135,37 @@ func TestTimestampString(t *testing.T) {
...
+		for pb.Next() {
+			if err := StdTimeUnmarshal(&t, data); err != nil {
+				panic(err)
+			}
+		}
+	})
+}
+
+func TestStdTimeUnmarshalAllocs(t *testing.T) {
+	data, err := StdTimeMarshal(time.Time{})
```
awalterschulze: how about using NewPopulatedTime, instead of an empty time?

types/timestamp_test.go
```diff
@@ -135,3 +135,37 @@ func TestTimestampString(t *testing.T) {
 func utcDate(year, month, day int) time.Time {
 	return time.Date(year, time.Month(month), day, 0, 0, 0, 0, time.UTC)
 }
+
+func BenchmarkStdTimeUnmarshal(b *testing.B) {
+	data, err := StdTimeMarshal(time.Time{})
```
awalterschulze: NewPopulated and maybe make a few in a slice, if possible?

Is RunParallel necessary?

awalterschulze:
Optional:
  - What about Duration
  - What about Marshaling
  - Add yourself to the CONTRIBUTORS or AUTHORS as appropriate

types/timestamp_test.go
bmizerany: What benefit do we get from randomly generated Times? What do you suggest we do with the slice of Times? RunParallel is not necessary and I'll remove.

types/timestamp_test.go
bmizerany: I'll add another AllocsPer with a time.Now() to prove allocs don't happen for both the zero value and non-zero.

bmizerany:
I think Duration, Marshaling, etc are better done in another PR to prevent this one from living too long. I'll be getting to those soon when my forcing function kicks in.

types/timestamp_test.go
bmizerany: I added a non-zero time to the bench and remove RunParallel.

bmizerany:
I believe I've address all of your concerns about the current patch. It would be nice to get this merged and I'm happy to send over Marshaling and Duration in sperate future PRs.

awalterschulze:
Ok maybe we can at least take on the Marshal of the timestamp in this commit as well.
It is doing `TimestampProto` twice.  It should be horribly inefficient.

awalterschulze:
hello?

timpalpant:
Can I pick this up? I'm in the same situation re: allocs due to StdTime fields. I will implement the marshaling and Duration equivalents if you're still ok with this approach.

PR #256, opened 2017-01-24
Fix unrendered markdown

Serabe:
Markdown inside HTML inside markdown is not rendered correctly. This fixes it inside extensions.md


awalterschulze:
Thanks

PR #258, opened 2017-02-07
Only import time when it is used

Serabe:
When using stdtime and typedecl options, an error is raised because we are
importing time and not using it (if the time field is not nullable).

This is fixed substituting `customImports` in the generator to receive a custom
struct containing three optional fields:

* *Alias* for the import
* *Path* for the import
* *UnderscorableSymbol* for an exported symbol in the package that can be used
as an expression (it will be used in `var _ = <Alias>.<UnderscorableSymbol>`)

Then, `writtenImports` is used to make sure we don't write the `var` assignment
twice.

If a `customImport` has no `UnderscorableSymbol`, then the statement won't be
generated at all.


awalterschulze:
Sorry about the delay.

Isn't there some way we could reuse the work done in generator/helper.go, the importedPackage type?

I would love not to import time when it is not used.
The only reason the others are imported is to stay close to the original golang/protobuf generated code.

Also can the test not be in the combos folder?
The combos folders are for switching marshaling on and off in various combinations.

Serabe:
I've been trying to migrate to `importedPackage` + `Use()` today and I still have the same problem as before. The `customImports` field is appended in `GoType` and this method returns the type as a string to be used. This string might not be used (depending on the case). `GoType` might be (and is being) called even if the type is not directly used in the output.

I am not able to distinguish when I am being called from a test file or the normal `.pb.go` file, thus, either generating it for both (not always needed) or for none (the test always needs it).

I'll be glad to change the test to another location. Which is the best place to put a problem that comes from using several plugins at once?

Thank you!

awalterschulze:
The folder could be anything in the root test folder so what about
`./test/declcustom` or `./test/issuexx` if you want to reference an issue.

protoc-gen-gogo/generator/helper.go
```diff
@@ -442,6 +442,6 @@ func (d *Descriptor) Path() string {
 
 func (g *Generator) useTypes() string {
```
awalterschulze: We should make sure that this only gets called, when we actually use types and we could make the same method for useTime.
Is more work really needed for other customtypes?

awalterschulze:
Ok so lets leave the use of importedPackage and see my comment in the review.

protoc-gen-gogo/generator/helper.go
Serabe: Not really.

Serabe:
@awalterschulze adapter your suggestions. Thank you for reviewing this!

protoc-gen-gogo/generator/generator.go
```diff
@@ -609,10 +609,33 @@ type Generator struct {
 	indent           string
 	writeOutput      bool
 
-	customImports  []string
+	customImports  []*customImport
 	writtenImports map[string]bool // For de-duplicating written imports
 }
 
+// customImport represents a custom import in the generated package
+type customImport struct {
```
awalterschulze: Do we still need this?
I would like to keep the diff with golang/protobuf in generator.go to a minimum.

test/issue260/README.md
```diff
@@ -0,0 +1,11 @@
+# The Bug
```
awalterschulze: I really like this file.  Good idea.

Serabe:
@awalterschulze done! Thank you for your quick review!

awalterschulze:
Perfect.  Thank you :)

Serabe:
Thank you!

PR #261, opened 2017-02-14
Fix cases where the fix for issue260 was not enough

Serabe:
The previous fix had some corner cases not covered by the `useTime` fix, since `time.Time` might only be used in the unmarhsaler plugin.

This PR makes `useTime` public and adds it to the lines of `unmarshaler` where it is needed.

Sorry, for having to do this fix.


awalterschulze:
Why is p.typesPkg.Use() not enough in the unmarshaler?

Serabe:
typesPkg is the gogoprotobuf types, I need to signal that I need the time package. This changes are not because of the line after but because of the line before.

awalterschulze:
Ok yes, my bad.
I am guessing creating a timesPkg might also have some issues conflicting with time when its already imported, but it might be worth exploring?

What about all the other plugins?

If timesPkg isn't an option, could we maybe let generator.UseTime() return a string "time"

Serabe:
I can do the second option. I review the rest of the plugins, but I was looking for uses of `tim.Time` or `time.Duration` and did not see anything besides the unmarshaler. I'll recheck tomorrow first thing in the morning.

Thank you!

awalterschulze:
No! Thank YOU very much.

Serabe:
My main problem is that `GoType` is used for several things at once. If I add the `useTime()` in there, I'll be calling it without knowing for sure that I'll be using it. The only thing I can think of is changing GoType to returning something like:

```go
type SingleType interface {
    Use() string
}

type singleType string

func (s *singleType) Use() string {
...
}

type singleTypeWithSideEffect struct {
    name string
    fn .     func()
}

func (s *singleType) Use() string {
    // Sideeffect
}
```

The refactor is quite big, but it would allow us to manage this deps in a better way.

awalterschulze:
I am busy making a simpler fix.

Basically only when time is imported I also add the `var _ = time.Kitchen` statement.
That way imports stay to a minimum except when typedecl=false and stdtime=true is used with nullable=false.

awalterschulze:
Sorry that I stepped in, but I don't like having things broken and it was quicker to fix it myself than to communicate.
I didn't know how I was going to fix it and trying things on the code myself also helped with getting the idea.
I also had to knock my head against the GoType function to see the real problem, like you originally said.
Thanks again for your help.
I really appreciate it.
And again sorry about how I handled this.

Serabe:
I have a fix for this case I'll be submitting tomorrow. But improving testing related to #262, I saw another problem with stdtime. More on it tomorrow, been a long day navigating `generator.go`

Serabe:
Rebasing...

johanbrandhorst:
:+1: for the work here, good job!

Serabe:
Ooook, so here it goes a summary of yesterday navigating protobuf.

# Initial case

My first attempt was to track where the import was lacking and making sure it would not happen again. The thing is that, for tracking that imports are done correctly, one cannot keep all the cases in one proto test since you can even have no individual case right but the proto as a whole can pass (more on this, later).

Some of the main problems with generating a message is that the `generateMessage` is long (~1K lines one method) and that the `GoType` method is the paradigm of coupling. `GoType` returns both a string with the typename and the wire type, and sometimes it gets called for the typename, sometimes for the wire type, and sometimes for both. That makes it pretty hard to add imports just based on resolving the typename because maybe the call is just interested on the wire type.

Then I considered applying the refactor I talked about here https://github.com/gogo/protobuf/pull/261#issuecomment-280045005 but changing typename from `string` to its own struct/interface is a hell (again, having a really long `generateMessage` does not help at all) and I even started implementing it but, sadly, it is too much and would required much more work than one could expect from it. It would also mean it is a pain to maintain changes coming from `golang/protobuf`.

Therefore, I went to a `useTime` approach just before inserting `time.X` in the code. Works.
I also added a `timePkg` struct to all plugins that might use `time`.

# The case of maps with stdtime/stdduration

When I started working on the fix, the first thing I did was splitting stdtype test in several tests. This helped me debugging the issue and showed a different issue with stdtime. The following paragraphs will be copypasted in a new issue.

If we have a field whose type is a map with `Timestamp` as its value type, and the `stdtime` is used in that field, then the imports are broken because...

1. [`GoMapType`](https://github.com/gogo/protobuf/blob/d2e1ade2d719b78fe5b061b4c18a9f7111b5bdc8/protoc-gen-gogo/generator/generator.go#L1919) gets called with the `FieldDescriptorProto` of the field itself.
2. In `GoMapType`, `GoType` is called for both the key and value types but...
3. Protobuf has no notion of types, but of fields, and therefore, bot the key and value ~types~ fields are not aware of the options in the map ~type~ field.

protoc-gen-gogo/generator/generator.go
```diff
@@ -2865,12 +2863,14 @@ func (g *Generator) generateMessage(message *Descriptor) {
 				val = "*c"
 			} else if gogoproto.IsStdTime(field) {
 				pkg := g.useTypes()
+				g.UseTime()
```
awalterschulze: Why do we need to UseTime twice?

protoc-gen-gogo/generator/generator.go
```diff
@@ -2879,12 +2879,14 @@ func (g *Generator) generateMessage(message *Descriptor) {
 				val = "c"
 			} else if gogoproto.IsStdDuration(field) {
 				pkg := g.useTypes()
+				g.UseTime()
 				g.P(`if err != nil {`)
 				g.In()
 				g.P(`return true, err`)
 				g.Out()
 				g.P(`}`)
 				g.P(`c := new(time.Duration)`)
```
awalterschulze: What about g.P(`c := new(`, g.UseTime(), `.time.Duration)`)

test/issue261/stdduration/map/nullable/issue261.proto
```diff
@@ -0,0 +1,39 @@
...
+syntax = "proto3";
+
+package issue261;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+import "google/protobuf/duration.proto";
+
+message MapStdTypes {
+  //map<int32,google.protobuf.Duration> duration = 4 [(gogoproto.stdduration) = true];
```
awalterschulze: Should this be commented?

awalterschulze:
There is a lot of tests here, which is great, but it also pointing to how complex is issue and can become.
There might be more cases we didn't think of.
Which scares me from a maintainability viewpoint.
Doesn't the current master still fix all these usecases?

The only downsides are:
  - it only causes an unnecessary import of time when `typedecl=false && (stdtime=true || stdduration=true)`
  - and it only causes a var _ statement when `stdtime=true || stdduration=true`

awalterschulze:
Great writeup of the issues by the way.

protoc-gen-gogo/generator/generator.go
Serabe: I think it was a problem with my rebase.

protoc-gen-gogo/generator/generator.go
Serabe: On it.

Serabe:
@awalterschulze I understand that it would be a nightmare to maintain but the bugs are still there no matter how. The best way would be to use `go imports` in the generated files. That would solve a lot of issues (though might open a few other ones). And no, current master does not fix all those use cases. The commented ones have been failing since the very beginning (you can check #266), they were just not tested. There will be many more cases. Yes, but these are the ones I caught.

Maybe a new kind of test that just generates lots and lots of single-field proto tests is needed. I don't know.

The code needs a huge refactor. And I don't mean gogo/protobuf but golang/protobuf. `GoType` is a nightmare on its own, `generateMessage` gives me the creeps, not having a notion of type and using fields as types in maps makes me shudder. I don't envy you and I am very thankful for your work. Without gogo/protobuf, [src-d/proteus](https://github.com/src-d/proteus) wouldn't have been possible at all. And I'm pretty sure that all those companies in the README are as thankful as me.

Please, feel free to drop this PR. None of my feelings will be hurt for sure. I went on with this so your comment [here](https://github.com/gogo/protobuf/pull/258#issuecomment-279134168) was a reality.

Proteus is working, which was my main concern. If the generated files have more imports than needed, it is no problem for me.

Thanks for reviewing all my PRs!

protoc-gen-gogo/generator/generator.go
```diff
@@ -2879,12 +2877,13 @@ func (g *Generator) generateMessage(message *Descriptor) {
 				val = "c"
 			} else if gogoproto.IsStdDuration(field) {
 				pkg := g.useTypes()
+				g.UseTime()
```
awalterschulze: UseTime is already called below, so this one is redundant

protoc-gen-gogo/generator/generator.go
```diff
@@ -2748,6 +2745,7 @@ func (g *Generator) generateMessage(message *Descriptor) {
 				} else if gogoproto.IsStdDuration(field) {
 					pkg := g.useTypes()
 					if gogoproto.IsNullable(field) {
+						g.UseTime()
```
awalterschulze: Do we need UseTime here?

test/issue261/stdduration/map/non-nullable/issue261.proto
```diff
@@ -0,0 +1,39 @@
...
+syntax = "proto3";
+
+package issue261;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+import "google/protobuf/duration.proto";
+
+message MapStdTypes {
+  //map<int32,google.protobuf.Duration> duration = 4 [(gogoproto.stdduration) = true, (gogoproto.nullable) = false];
```
awalterschulze: Should this be commented?

test/issue261/stdduration/map/nullable/issue261.proto
awalterschulze: ?

test/issue261/stdtime/map/non-nullable/issue261.proto
```diff
@@ -0,0 +1,39 @@
...
+syntax = "proto3";
+
+package issue261;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+import "google/protobuf/timestamp.proto";
+
+message MapStdTypes {
+  //map<int32,google.protobuf.Timestamp> timestamp = 2 [(gogoproto.stdtime) = true, (gogoproto.nullable) = false];
```
awalterschulze: Should this be commented?

test/issue261/stdtime/map/nullable/issue261.proto
```diff
@@ -0,0 +1,39 @@
...
+syntax = "proto3";
+
+package issue261;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+
+import "google/protobuf/timestamp.proto";
+
+message MapStdTypes {
+  //map<int32,google.protobuf.Timestamp> timestamp = 2 [(gogoproto.stdtime) = true];
```
awalterschulze: Should this be commented?

Serabe:
Fix what I could fix. The commented tests have never been working, not even before I started contributing.

The problem is that in a map, the keys and values are not represented by `Type`s but by `FieldDescriptor`s. Therefore, when in `(g *Generator) GoMapType` `(g *Generator) GoType` is called with the `FieldDescriptor`s for the key type and the value type, the options are not available since they are different fields. Therefore, the `GoType` method has no access to the stdtime nor stdduration defined in the actual field defined in the proto, not the ones defined to be used as key type or value types in the map.

Serabe:
Rebased

awalterschulze:
Sorry for the delay.
This is quite complex.
You are right, this has always been broken, even before typedecl was added, since it has nothing to do with it.
I'll look into this as soon as I can.
But today isn't good :(

Serabe:
NP

awalterschulze:
This was really hard to find, but this should fix it
https://github.com/gogo/protobuf/commit/f1756eb6594909c6b25c3f4be28a127aa16b28a7#diff-e0188cb7643b9790acd32abc395a5659

awalterschulze:
False alarm :(

awalterschulze:
This should be better https://github.com/gogo/protobuf/commit/83faaee7bbbdf0c59310ec67d340bc0cbe77053f

awalterschulze:
So I think all the bugs that this pull request is pointing out has been fixed.
Or am I still missing something?

Serabe:
Rebased your changes.

awalterschulze:
And I see all the maps are uncommented as well.
So everything works now :)

Serabe:
I am going to create a new extension for making nullable the value types in a map. Can I close this or are you still interested on something from here?

Thank you!

awalterschulze:
Yes I think you can close this.

I am interested in the new usecase with maps?  Currently I am holiday for a few weeks.

Serabe:
Some people needs to represent `map[Something]*OtherThing`, so I was planning to add an option for that in gogo/protobuf.

awalterschulze:
Isn't that already possible for messages or what is Otherthing

Serabe:
Sorry, it was a typo `map[Something]OtherThing`.

awalterschulze:
Doesn't nullable false work?

Serabe:
It does, thank you!

Serabe:
Reopening this since we have found a similar bug again. I'll dive tomorrow or next week to see what happened, if was something that I did not test or a regression.

Thank you!

awalterschulze:
hello?

Serabe:
Closing this as I don't use protobuf anymore. Thank you!

Issue #222, opened 2016-10-25
Using protosize_all and sizer_all with a oneof generates invalid code


johanbrandhorst:
Here's a gist that reproduces the problem:
https://gist.github.com/johanbrandhorst/880326d70e69d191d7560fd498031605

Essentially, the interface is created with the `ProtoSize` method, but it is not implemented, presumably because there is some check to prevent both `ProtoSize` and `Size` methods to be generated. I would suggest a compile time warning here if possible.

``` go
type isValue_Type interface {
    isValue_Type()
    Size() int
    ProtoSize() int
}
```

awalterschulze:
Yes a generation time warning should already exist, this is weird.

Something here would be nice
https://github.com/gogo/protobuf/blob/master/plugin/size/size.go#L581

That throws an error, not just a warning, like this
https://github.com/gogo/protobuf/blob/master/plugin/defaultcheck/defaultcheck.go#L116

johanbrandhorst:
Cool, I'll take a look, thanks for the pointers.

awalterschulze:
You still want to take a look?

johanbrandhorst:
Sorry, I got sidetracked on this one, I'll take another look! :100:

PR #271, opened 2017-03-07
Compile time warning when both Sizer and Protosizer are used

johanbrandhorst:
Fixes issue #222

### DIFF
--- a/plugin/size/size.go
+++ b/plugin/size/size.go
@@ -121,6 +121,7 @@ package size
 
 import (
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


awalterschulze:
Perfect :)

awalterschulze:
Ah wait no.

This still requires an entry in the top level Makefile to actually call the test.

johanbrandhorst:
I thought about that, but it will be called by `go test ./...`, so it shouldn't be needed?

awalterschulze:
Ah my bad.
Context switching between projects can be tiring.
Thank you for keeping this sane :)

Issue #277, opened 2017-03-29
For `moretags` option, how can I make it like `jsontag `, automatically add the camel key and omitempty


lnshi:
@awalterschulze Thank you very much for making this awesome repo and actively maintain it.

Now I am using the `moretags` feature for `bson`, just like you demoed here: https://github.com/gogo/protobuf/blob/a11c89fbb0ad4acfa8abc4a4d5f7e27c477169b1/test/tags/tags.proto 

It behaves exactly as what you described, but I am thinking how to make it just like the `jsontag`? I mean automatically generate the camel-case key and the `omitempty` tag.

---- Edited
I just submitted the PR, pls help to review, feel free to require any changes.

awalterschulze:
I would recommend to rather create a vanity binary
for example
https://github.com/gogo/protobuf/blob/master/protoc-gen-gogofast/main.go

And then for each field add a moretags extension to that field with the camelCased and omitempty bson tag text, just like you would have manually done with moretags.

What do you think?

lnshi:
Ah, definitely better than my approach, I don't know that functionality.

But haven't exactly understand how will it work, roughly checked a bit, so:
```
// ExtensionDesc represents an extension specification.
// Used in generated code from the protocol compiler.
type ExtensionDesc struct {
	ExtendedType  Message     // nil pointer to the type that is being extended
	ExtensionType interface{} // nil pointer to the extension type
	Field         int32       // field number
	Name          string      // fully-qualified name of extension, for text formatting
	Tag           string      // protobuf tag style
	Filename      string      // name of the file in which the extension is defined
}
```

Then I can refer this func in https://github.com/gogo/protobuf/blob/a11c89fbb0ad4acfa8abc4a4d5f7e27c477169b1/vanity/field.go
```
func SetBoolFieldOption(extension *proto.ExtensionDesc, value bool) func(field *descriptor.FieldDescriptorProto) {
	return func(field *descriptor.FieldDescriptorProto) {
		if FieldHasBoolExtension(field, extension) {
			return
		}
		if field.Options == nil {
			field.Options = &descriptor.FieldOptions{}
		}
		if err := proto.SetExtension(field.Options, extension, &value); err != nil {
			panic(err)
		}
	}
}
```

but just return a `func` to change the `extension.Options` instead of the `field.Tag`? Am I correct?

awalterschulze:
I think you want something like this

```
func SetBsonTagFieldOption(field *descriptor.FieldDescriptorProto) {
[...]

PR #279, opened 2017-03-29
Address issue #277

lnshi:


awalterschulze:
So the idea is that you create this as your own vanity binary in your own repo.
Its kind of hard to decide which vanity binaries to create and what combination of options they will contain.  For example lots of people don't use bson. But if you create your own binary then you have the freedom to do exactly what you want.  And it should be pretty easy to maintain.  I mean there is no reason why this cannot all live in one main.go file that simply imports gogoprotobuf.

lnshi:
Ah, sorry, I guess I just spammed ur email, sorry, I am gonna close this. Just because I found if I want to run the tests against my fork, I need to change lots of import paths...

Understand, but can we consider to make this a command option? so people can choose there....

lnshi:
And @awalterschulze, can u do me a bit more favour, I just cannot figure out why the last test sets always fails, checked the logs, just cannot find helpful info...

awalterschulze:
govet is saying the struct tag's syntax is broken, but I can't spot the problem
```
test/fuzztests/fuzz.pb.go:50: struct field tag `protobuf:"fixed64,1,rep,packed,name=Field1,json=field1" json:"Field1,omitempty" bson: "Field1,omitempty"` not compatible with reflect.StructTag.Get: bad syntax for struct tag value
```
And then you just need to regenerate all the code, `make regenerate` for protoc 3.2.0

awalterschulze:
At some point you are going to need your own test setup :)

awalterschulze:
That bash `if` means that only for go 1.8 and protoc 3.1.0 make sure that the generated code is the code that is check it by doing a git diff after the code has been regenerated.

awalterschulze:
I am can put a link from my documentation to your protoc binary on github for people who want bson encoding.  What do you think?

lnshi:
@awalterschulze how come only for that test case there is one extra space coming after the bson:

awalterschulze:
Which test are we talking about?

PR #281, opened 2017-03-29
Optimize uint32writer

cpuguy83:
Don't use `binary.Write` which makes 2 allocs. Instead since we know we always want a uint32 encode this directly with `binary.PutUint32`, which is what `binary.Write` does.

Creates a fast path and a slow path. Fast path is for types that satisfy `marshaller` where we can work with just a single buffer.
Slow path is for everything else where it uses `proto.Marshal`.

```
benchmark                                  old ns/op     new ns/op     delta
BenchmarkUint32DelimWriterMarshaller-8     159           77.8          -51.07%
BenchmarkUint32DelimWriterFallback-8       879           764           -13.08%

benchmark                                  old allocs     new allocs     delta
BenchmarkUint32DelimWriterMarshaller-8     3              0              -100.00%
BenchmarkUint32DelimWriterFallback-8       9              7              -22.22%

benchmark                                  old bytes     new bytes     delta
BenchmarkUint32DelimWriterMarshaller-8     80            0             -100.00%
BenchmarkUint32DelimWriterFallback-8       728           712           -2.20%
```

### DIFF
--- a/io/uint32.go
+++ b/io/uint32.go
@@ -30,53 +30,65 @@ package io
 
 import (
 	"encoding/binary"
-	"github.com/gogo/protobuf/proto"
 	"io"
+
+	"github.com/gogo/protobuf/proto"
 )
 
+const uint32BinaryLen = 4
+
 func NewUint32DelimitedWriter(w io.Writer, byteOrder binary.ByteOrder) WriteCloser {
-	return &uint32Writer{w, byteOrder, nil}
+	return &uint32Writer{w, byteOrder, nil, make([]byte, uint32BinaryLen)}
 }
 
 func NewSizeUint32DelimitedWriter(w io.Writer, byteOrder binary.ByteOrder, size int) WriteCloser {
-	return &uint32Writer{w, byteOrder, make([]byte, size)}
+	return &uint32Writer{w, byteOrder, make([]byte, size), make([]byte, uint32BinaryLen)}
 }
 
 type uint32Writer struct {
 	w         io.Writer
 	byteOrder binary.ByteOrder
 	buffer    []byte
+	lenBuf    []byte
 }
 
-func (this *uint32Writer) WriteMsg(msg proto.Message) (err error) {
-	var data []byte
-	if m, ok := msg.(marshaler); ok {
-		n, ok := getSize(m)
-		if !ok {
-			data, err = proto.Marshal(msg)
-			if err != nil {
-				return err
-			}
-		}
-		if n >= len(this.buffer) {
-			this.buffer = make([]byte, n)
-		}
-		_, err = m.MarshalTo(this.buffer)
-		if err != nil {
-			return err
-		}
-		data = this.buffer[:n]
-	} else {
-		data, err = proto.Marshal(msg)
-		if err != nil {
-			return err
-		}
+func (this *uint32Writer) writeFallback(msg proto.Message) error {
+	data, err := proto.Marshal(msg)
+	if err != nil {
+		return err
 	}
+
 	length := uint32(len(data))
-	if err = binary.Write(this.w, this.byteOrder, &length); err != nil {
+	this.byteOrder.PutUint32(this.lenBuf, length)
+	if _, err = this.w.Write(this.lenBuf); err != nil {
 		return err
 	}
 	_, err = this.w.Write(data)
+	return err
+}
+
+func (this *uint32Writer) WriteMsg(msg proto.Message) error {
+	m, ok := msg.(marshaler)
+	if !ok {
+		return this.writeFallback(msg)
+	}
+
+	n, ok := getSize(m)
+	if !ok {
+		return this.writeFallback(msg)
+	}
+
+	size := n + uint32BinaryLen
+	if size > len(this.buffer) {
+		this.buffer = make([]byte, size)
+	}
+
+	this.byteOrder.PutUint32(this.buffer, uint32(n))
+	if _, err := m.MarshalTo(this.buffer[uint32BinaryLen:]); err != nil {
+		return err
+	}
+
+	_, err := this.w.Write(this.buffer[:size])
 	return err
 }
 
--- a/io/uint32_test.go
+++ b/io/uint32_test.go
@@ -1,0 +1,38 @@
+package io_test
+
+import (
+	"encoding/binary"
+	"io/ioutil"
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/gogo/protobuf/test"
+	example "github.com/gogo/protobuf/test/example"
+
+	"github.com/gogo/protobuf/io"
+)
+
+func BenchmarkUint32DelimWriterMarshaller(b *testing.B) {
+	w := io.NewUint32DelimitedWriter(ioutil.Discard, binary.BigEndian)
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	msg := example.NewPopulatedA(r, true)
+
+	for i := 0; i < b.N; i++ {
+		if err := w.WriteMsg(msg); err != nil {
+			b.Fatal(err)
+		}
+	}
+}
+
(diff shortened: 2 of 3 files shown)


io/uint32.go
```diff
@@ -72,11 +79,19 @@ func (this *uint32Writer) WriteMsg(msg proto.Message) (err error) {
 			return err
 		}
 	}
+
 	length := uint32(len(data))
-	if err = binary.Write(this.w, this.byteOrder, &length); err != nil {
+	this.byteOrder.PutUint32(this.lenBuf[:uint32BinaryLen], length)
```
awalterschulze: Why is this uint32BinaryLen needed here, the lenBuf is already the fixed size?

io/uint32.go
```diff
@@ -72,11 +79,19 @@ func (this *uint32Writer) WriteMsg(msg proto.Message) (err error) {
...
+	if _, err = this.writeBuf.Write(this.lenBuf[:uint32BinaryLen]); err != nil {
```
awalterschulze: same here

io/uint32_test.go
```diff
@@ -0,0 +1,33 @@
...
+	defer os.Remove(f.Name())
+	defer f.Close()
+
+	w := io.NewUint32DelimitedWriter(f, binary.BigEndian)
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	msg := test.NewPopulatedA(r, true)
+
+	for i := 0; i < b.N; i++ {
+		if err := w.WriteMsg(msg); err != nil {
+			panic(err)
```
awalterschulze: b.Fatal

awalterschulze:
Couldn't the slow stream problem also be solved by the user with the slow stream passing in a buffered writer into the NewDelimitedWriter?

awalterschulze:
Good catch at binary.Write :)

io/uint32_test.go
cpuguy83: I did panic because other tests were using it, will definitely change over.

cpuguy83:
@awalterschulze Yes, the caller could pass in a buffer but I don't think this would improve anything since the writes are opaque at this point.

cpuguy83:
Updated

awalterschulze:
Can we try some bigger structs in the benchmark.
I still can't believe that copying the buffer of the message twice instead of once is faster.

So what you mean with opaque is that the bufferedwriter would not be able to tell which writes are length writes and which are marshaled buffer writes and so not be able to deduce the most efficient logic?

cpuguy83:
It would be nice to have a bigger proto that implements `MarshalTo` since there's really two paths here.
But in any case, changed this around a little bit so it doesn't even need the 2nd buffer.
Now if the past in proto meets all of these, it will have no allocations:

- Implements `MarshalTo()`
- `getSize()` works
- Passed in buffer is smaller than the internal buffer

This is the fast path.
If the passed in proto is larger than the internal buffer than it will require an alloc.

The slow path's allocations is reduced by 2.
Only 1 allocation is required for the slow path except for anything that is incurred by the actual marshaller.

Not a huge fan of how this can store larger and larger buffers forever, but that is the existing behavior.

See the original comment for updated benchmark results.

awalterschulze:
All the thetest.proto structs with a MarshalTo are available in ./test/combos/both

io/uint32.go
```diff
@@ -30,56 +30,69 @@ package io
...
+
+	size := n + uint32BinaryLen
+	if size > len(this.buffer) {
+		this.buffer = make([]byte, size)
+	}
+	if _, err := m.MarshalTo(this.buffer[uint32BinaryLen:]); err != nil {
+		return err
+	}
+
+	this.byteOrder.PutUint32(this.buffer, uint32(n))
```
awalterschulze: You already have the size so you can put it first in the natural order.

awalterschulze:
It would be nice to compare the same struct for fallback vs non fallback.

awalterschulze:
Except for these comments it looks really good.

awalterschulze:
The uint32Reader has a maxSize to avoid the problem of the buffer growing ever larger.
But maybe lets tackle that in a separate commit, because it will change the api or at least add another extra constructor.

cpuguy83:
@awalterschulze Can you explain how I can use these without causing panics?

```
<truncated>
2017/04/05 08:42:51 proto: duplicate proto type registered: test.NinOptNonByteCustomType
2017/04/05 08:42:51 proto: duplicate proto type registered: test.NidRepNonByteCustomType
2017/04/05 08:42:51 proto: duplicate proto type registered: test.NinRepNonByteCustomType
2017/04/05 08:42:51 proto: duplicate proto type registered: test.ProtoType
panic: proto: duplicate enum registered: test.TheTestEnum

goroutine 1 [running]:
github.com/gogo/protobuf/proto.RegisterEnum(0x1328308, 0x10, 0xc420019200, 0xc420019230)
        /Users/bgoff/go/src/github.com/gogo/protobuf/proto/properties.go:907 +0x2ac
github.com/gogo/protobuf/test/combos/both.init.1()
        /Users/bgoff/go/src/github.com/gogo/protobuf/test/combos/both/thetest.pb.go:1505 +0xc8a
github.com/gogo/protobuf/test/combos/both.init()
        /Users/bgoff/go/src/github.com/gogo/protobuf/test/combos/both/uuid.go:134 +0x88f
github.com/gogo/protobuf/io_test.init()
        /Users/bgoff/go/src/github.com/gogo/protobuf/io/uint32_test.go:69 +0x76
main.init()
        github.com/gogo/protobuf/io/_test/_testmain.go:66 +0x58
```

awalterschulze:
Ah damn, I forgot about stupid Register functions. Lets leave the benchmarking of the same struct :(
I really want to make a plugin that just does not generate the code to Register any of the messages.

cpuguy83:
There seems to be something breaking here not covered by tests. I'll look into it.

cpuguy83:
Ok, nevermind, it's not this. 😅

cpuguy83:
Went ahead and changed the ordering so `PutUint32` is before writing the proto to the buffer.

io/uint32.go
```diff
@@ -30,56 +30,69 @@ package io
...
-	return &uint32Writer{w, byteOrder, make([]byte, size)}
+	return &uint32Writer{w, byteOrder, make([]byte, size), make([]byte, uint32BinaryLen), proto.NewBuffer(nil)}
 }
 
 type uint32Writer struct {
 	w         io.Writer
 	byteOrder binary.ByteOrder
 	buffer    []byte
+	lenBuf    []byte
+	protoBuf  *proto.Buffer
```
awalterschulze: Where is this used?

cpuguy83: whoops

awalterschulze:
The rest looks good :)
Also would you like to add yourself as an AUTHOR or CONTRIBUTOR whichever is appropriate?

io/uint32.go
```diff
@@ -30,56 +30,68 @@ package io
...
 	"io"
+
+	"github.com/gogo/protobuf/proto"
 )
 
+const uint32BinaryLen = 4
+
 func NewUint32DelimitedWriter(w io.Writer, byteOrder binary.ByteOrder) WriteCloser {
-	return &uint32Writer{w, byteOrder, nil}
+	return &uint32Writer{w, byteOrder, nil, make([]byte, uint32BinaryLen), proto.NewBuffer(nil)}
```
awalterschulze: The proto.NewBuffer(nil) is still here and below

awalterschulze:
If checks pass I'll merge :)

awalterschulze:
Thank you very much :)  Really cool optimization.

cpuguy83:
Thanks, sorry for the back and forth there!

awalterschulze:
No problem that's how it usually goes ☺️

Issue #13, opened 2014-12-03
Known Issue: unknown behaviour for proto.Equal


awalterschulze:
_From [awalterschulze](https://code.google.com/u/awalterschulze/) on May 07, 2014 21:17:58_

proto.Equal(a, b Message) bool has not been tested with all gogoprotobuf scenarios.

If you are using this please comment so I can up the priority.

_Original issue: http://code.google.com/p/gogoprotobuf/issues/detail?id=13_

dennwc:
Found an issue with this combination of options:
```proto
message M {
    repeated bytes arr = 1 [(gogoproto.customtype) = "MyType", (gogoproto.nullable) = false];
}
```
It will generate next code for M:
```Go
type M struct {
    Arr []MyType `protobuf:"bytes,1,rep,name=arr,customtype=MyType" json:"arr"`
}

func (this *M) Equal(that interface{}) bool {
    // ...
    that1, ok := that.(*M)
    if !ok {
        return false
    }
    // ...
    if !this.Arr[i].Equal(that1.Arr[i]) { // should be &that1.Arr[i]
        return false
    }
    // ...
}
```
It will fail because `Equal` method will try to cast a parameter to `*MyType`.

awalterschulze:
This should be reported in another issue, but you could check out in test/thetest.proto
```
message NidRepCustom {
	repeated bytes Id = 1 [(gogoproto.customtype) = "Uuid", (gogoproto.nullable) = false];
	repeated bytes Value = 2 [(gogoproto.customtype) = "github.com/gogo/protobuf/test/custom.Uint128", (gogoproto.nullable) = false];
}
```
The equal method does the following
```
func (uuid Uuid) Equal(other Uuid) bool {
	return bytes.Equal(uuid[0:], other[0:])
}
```
Does this help?

dennwc:
Yes, it will definitely would, but this will not work for auto-generated "custom" type. I want to point out, that behavior of this function is not symmetric: it will pass `Arr[i]`, but auto-generated variant of `Equal` always assumes a pointer to the type. So maybe it should pass pointers always, or as an alternative, auto-generated `Equal` functions should check for a non-pointer argument also.

awalterschulze:
Could you please create a new issue for this?  Before we start having a longer conversation.

andreimatei:
I've run into problems with `proto.Equal()` and a proto where we use `gogoproto.customtype` for a `bytes` field. I hope this is the right place to report.

This is the proto:
[...]

PR #285, opened 2017-04-20
Implement EqualIgnoreOrder

swelly127:
Hey I work at Dropbox, I think this function would be helpful for a lot of people who need to test equalness in proto objects when they don't care about order of elements in slices.

### DIFF
--- a/proto/equal.go
+++ b/proto/equal.go
@@ -45,6 +45,7 @@ Equal returns true iff protocol buffers a and b are equal.
 The arguments must both be pointers to protocol buffer structs.
 
 Equality is defined in this way:
+ For Equal:
   - Two messages are equal iff they are the same type,
     corresponding fields are equal, unknown field sets
     are equal, and extensions sets are equal.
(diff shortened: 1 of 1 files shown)


awalterschulze:
The first problem here is that gogoprotobuf does not fully support proto.Equal
https://github.com/gogo/protobuf/issues/13
It has not been tested enough and I am sure there are edge cases, which don't work.
I recommend generating an Equal method using gogoprotobuf's extensions or you are welcome to help me fix proto.Equal by adding more tests.

The second problem is that this code is part of the fork with golang/protobuf.  I don't suspect they will go for such a pull request, which means I will have to support this diff and with every merge from golang/protobuf.  That means that I will want this feature to be something that a lot of people want.

Maybe a simpler way to solve this is to write a function that uses reflect and recursively sorts all fields that are slices and then does a reflect.DeepEqual after that?
Or if this is only for a specific case, maybe you don't even need to use reflect.

awalterschulze:
I am closing this, but just comment if you would like it reopened.

PR #291, opened 2017-05-12
Support for multiple named grpc service instances

simonferquel:
At Docker we have a particular case where we would like to expose multiple implementations of the same service interface on a single grpc server.

To do that, we need a way to register the grpc service by giving it a custom name (instead of the generated service name), and we need a client proxy that can dynamically change its methods / streams address depending on this custom name.

This PR add a custom parameter that modify codegen to support that.

So if you pass arguments to protoc like `--gogoslick_out=plugins=grpc,grpc-named-instance=true:.`, in addition to the normal service client / server interfaces and proxies, it will create the following functions:
 - `func RegisterNamed<servicename>Server(s *grpc.Server, srv <servicename>Server, name string)` : register a named service instance
- `func NewNamed<servicename>Client(cc *grpc.ClientConn, name string) <servicename>Client` : that will create a named proxy (interface implementation is slightly different so that it generates method urls dynamically, depending on the passed name)


awalterschulze:
We usually use extensions to add extra code generation, instead of command
line flags. But first

Would you consider explaining your usecase a bit more. I would like to
understand how other users might also want to use this feature.

Also maybe this would be better as a different decoupled code generation
plugin?

simonferquel:
@awalterschulze First some context:
At Docker we are currently working on using grpc to expose client-side services to the daemon. What we do, is that on some service endpoints and under some circonstance (minimal API version supported by both client and daemon), we hijack the http connection and expose a grpc server from the client to be consumed by the daemon.

This opens up scenarios like:
- delegated authentication: on docker build, the daemon parses a dockerfile, and for each referenced image, asks the client for an Access token with just enough rights to pull that specific image
- file system caching: instead of sending the whole build context directory each time we issue a Docker Build, client and server negociate a diff using a grpc streaming service
- ssh tunneling: client can be used as an ssh tunneling proxy so that the daemon can access external resources trough SSH without having to obtain the actual credentials
And this is just the beginning of our thoughts about it.

There are 2 main reasons for which we want to have multiple instances of the same service (or multiple different implementations of the same service) exposed on this particular grpc server:
- Versioning: We have no control of which particular version of both client and daemon will be used. Sometimes they will be in sync, but sometimes client or daemon will be more up-to-date. So we want to be able to expose multiple versions of the same services, in cases where the public signature of the service does not change, but the actual behavior change can cause issues. So what we want here is to be able to have n different implementations of the same services exposed with service names "myservice.1, myservice.2, ... myservice.n"
- Configured instances: For the file system caching scenario for example, we would like to be able to expose multiple `FSCacheService` instances configured with different root directories. That would open things like more composite `docker build` scenarios with multiple different client context directories synchronized independently with the daemon side cache.

Now, on the implementation side, I could re-engineer as a new extension. The thing is that in this case I see it more like a customization of the generated grpc-related code than a completely new addition (and to avoid code duplication I found it quite easier to just modify the current grpc plugin). Anyway if you confirm you prefer it as an extension I'll refactor it and amend the PR.

awalterschulze:
The more I look at this, the more I think a separate plugin would be better.
It does require some code duplication, but its only 463 lines of which some can be deleted.
The grpc plugin is also one of the simplest plugins, so it should be easy to maintain.
What you loose in code duplication you gain in flexibility.

simonferquel:
Ok, I'll write a duplicate plugin so. How would you name it grpc_with_custom_names ? (seems a bit long...)

awalterschulze:
Cool.  I think its at least worth to see what it looks like, I mean maybe I
am wrong, but I would like to see.

How about namedgrpc or grpcnamed.

PR #292, opened 2017-05-17
Added support for named instances with a grpnamed plugin

simonferquel:
This is a followup of #291 using a separate grpcnamed plugin.
Code generation is a bit modified to be able to host multiple named instances of the same grpc server interface.

### DIFF
--- a/protoc-gen-gogo/grpcnamed/grpc.go
+++ b/protoc-gen-gogo/grpcnamed/grpc.go
@@ -1,0 +1,480 @@
+// Go support for Protocol Buffers - Google's data interchange format
+//
+// Copyright 2015 The Go Authors.  All rights reserved.
+// https://github.com/golang/protobuf
+//
+// Redistribution and use in source and binary forms, with or without
+// modification, are permitted provided that the following conditions are
+// met:
+//
+//     * Redistributions of source code must retain the above copyright
+// notice, this list of conditions and the following disclaimer.
+//     * Redistributions in binary form must reproduce the above
+// copyright notice, this list of conditions and the following disclaimer
+// in the documentation and/or other materials provided with the
+// distribution.
+//     * Neither the name of Google Inc. nor the names of its
+// contributors may be used to endorse or promote products derived from
+// this software without specific prior written permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+// "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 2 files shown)


simonferquel:
@awalterschulze I have no idea how to create a test for this one though...

protoc-gen-gogo/grpcnamed/grpc.go
```diff
@@ -0,0 +1,480 @@
...
+	g.P("Metadata: \"", file.GetName(), "\",")
+	g.P("}")
+	g.P("}")
+	g.P()
+	// Server registration.
+	g.P("func RegisterNamed", servName, "Server(s *", grpcPkg, ".Server, srv ", serverType, ", name string) {")
+	g.P("s.RegisterService(_named", servName, "ServiceDesc(name), srv)")
+	g.P("}")
+	g.P()
+	g.P("func Register", servName, "Server(s *", grpcPkg, ".Server, srv ", serverType, ") {")
```
awalterschulze: I thought that this plugin would only generate the RegisterNamed functions and that you won't need to generate the other Register functions, since they will be generated by the original grpc plugin.
This plugin should only need to do the extra code generation, which means it should require much less code.

simonferquel: The problem is that if I do it as a plugin, I can't be sure that the grpc plugin is enabled as well. So I decided to implement the whole thing. But the non-named version can be easily removed (they are only passing default named values to the NewNamed<..> functions.

awalterschulze: I think its ok to assume that the grpc plugin was also used.  minor Documentation and a golang Compiler error should be enough to help most users.

awalterschulze: I do the same with the marshal plugin which assumes that the size plugin was also enabled.

simonferquel: I made the grpcnamed plugin additive to grpc output. In parallel I've made a concurrent PR with named instance generation enabled trough a protobuf extension.

I kind of prefer the plugin based thing, as it does not require to mess with protoc import paths.

protoc-gen-gogo/grpcnamed/grpc.go
```diff
@@ -0,0 +1,377 @@
...
+	g.P(grpcPkg, ".ServerStream")
+	g.P("}")
+	g.P()
+
+	g.P("type ", streamType, " struct {")
+	g.P(grpcPkg, ".ServerStream")
+	g.P("}")
+	g.P()
+
+	if genSend {
```
awalterschulze: Do you still need all of this?

awalterschulze:
I would test this be at least having a proto file which each type of grpc service.

PR #293, opened 2017-05-17
Named grpc instances as ext

simonferquel:
This is a followup of #291 using a protobuf extension handled by the grpc plugin.
Setting option `goproto_named_instances` true triggers generation of code allowing to expose or consume named instances of a given service.

This is an alternative to #292 
cc @awalterschulze


awalterschulze:
Yes I think we are past using the extension now.  We can use it like the grpc plugin on the command line, since it is so totally decoupled now :)

awalterschulze:
Closing because of inactivity, but simply make a comment if you want it reopened.

PR #302, opened 2017-06-27
protoc-gen-gogo: int64 and uint64 types use string json option

jdef:
int64 and uint64 types are marshaled by jsonpb as string types and should be
tagged as `string` for interop with other golang json packages.

### DIFF
--- a/protoc-gen-gogo/generator/generator.go
+++ b/protoc-gen-gogo/generator/generator.go
@@ -2113,6 +2113,12 @@ func (g *Generator) generateMessage(message *Descriptor) {
 			if !gogoproto.IsNullable(field) && !repeatedNativeType {
 				jsonTag = jsonName
 			}
+			// for compat w/ jsonpb marshal/unmarshal
+			switch *field.Type {
+			case descriptor.FieldDescriptorProto_TYPE_INT64,
+				descriptor.FieldDescriptorProto_TYPE_UINT64:
+				jsonTag += ",string"
+			}
 			gogoJsonTag := gogoproto.GetJsonTag(field)
 			if gogoJsonTag != nil {
 				jsonTag = *gogoJsonTag


awalterschulze:
I think this is part of https://github.com/golang/protobuf/issues/256
Basically json and jsonpb are not compatible and there is lots of work to be done to make it so.

jdef:
OK, in that case is there a way to disable `jsonpb` from the tests generated by protoc-gen-gogo? otherwise I end up with broken tests because `jsonpb` marshals JSON in a way that's incompatible w/ respect to the json tags declared for the gogo-generated structs.

jdef:
I'm using `testgen` but not `jsontag` .. and the generated json tests don't work if I write or generate my own JSON unmarshaler (e.g. via ffjson) for the generated struct types. The configuration I'm using **used to work** with an older version of gogo but when I sync'd to master today the generated tests broke because of this

awalterschulze:
Ok so you are using `testgen` and ffjson.
So ffjson is probably confusing jsonpb

jdef:
Actually jsonpb is confusing ffjson because it's generating string values for int64/uint64 fields. and ffjson unmarshalling explodes because of this

awalterschulze:
Ok maybe its because I am sick, but I am struggling to understand you.
This is an example test generated by testgen

```

func TestNinOptNativeJSON(t *testing.T) {
	seed := time.Now().UnixNano()
	popr := math_rand.New(math_rand.NewSource(seed))
	p := NewPopulatedNinOptNative(popr, true)
	marshaler := github_com_gogo_protobuf_jsonpb.Marshaler{}
	jsondata, err := marshaler.MarshalToString(p)
	if err != nil {
		t.Fatalf("seed = %d, err = %v", seed, err)
	}
	msg := &NinOptNative{}
	err = github_com_gogo_protobuf_jsonpb.UnmarshalString(jsondata, msg)
	if err != nil {
		t.Fatalf("seed = %d, err = %v", seed, err)
	}
	if err := p.VerboseEqual(msg); err != nil {
		t.Fatalf("seed = %d, %#v !VerboseProto %#v, since %v", seed, msg, p, err)
	}
	if !p.Equal(msg) {
		t.Fatalf("seed = %d, %#v !Json Equal %#v", seed, msg, p)
	}
}
```

jsonpb is marshaling and int64 field to a string.
Then jsonpb is used to unmarshal.
Where is ffjson coming into play?

jdef:
https://github.com/gogo/protobuf/blob/master/jsonpb/jsonpb.go#L675

lots of calls to `json.Unmarshal` in there. for example: https://github.com/gogo/protobuf/blob/master/jsonpb/jsonpb.go#L906

ffjson decorates gogo-generated struct types with json unmarshal funcs that are executed in situations like this.

i skimmed the jsonpb marshal code and it's definitely wrapping int64/uint64 types with quotes to stringify them. i don't know why jsonpb isn't invoking ffjson's custom marshal func - feels like dissonance to NOT invoke a possibly custom marshal func, but then check for a possibly customized unmarshal func.

as long as jsonpb is going to possibly unvoke golang's `json.Unmarshal` func (which checks for custom unmarshal implementatins) then gogo-protoc should be properly json-tagging fields of generated structs.

awalterschulze:
Ok thats a good point.
I am going to wait till I feel better before I answer again.
But at least now I have more information.
Thank you

awalterschulze:
All json.Unmarshal calls are done one json.RawMessage, which shouldn't be a problem.
Except for these two calls, which are specific to gogoprotobuf:

This one you pointed out is probably the problem:
https://github.com/gogo/protobuf/blob/master/jsonpb/jsonpb.go#L906

Only on byte slices, so it is probably not the problem:
https://github.com/gogo/protobuf/blob/master/jsonpb/jsonpb.go#L1000-L1023

There is one more call to json.Unmarshal, but it is also present in golang/protobuf and so I guess this is not the problem.
https://github.com/gogo/protobuf/blob/master/jsonpb/jsonpb.go#L1088

awalterschulze:
So lets focus on https://github.com/gogo/protobuf/blob/master/jsonpb/jsonpb.go#L906

It is there for custom struct types.
So we should try and distinguish between customtype structs and structs which have generated UnmarshalJSON methods.

awalterschulze:
I think the first we need is a test to show how this is currently broken.  Can you provide that?

awalterschulze:
Here is an example customtype in question
https://github.com/gogo/protobuf/blob/master/test/t.go

We can see that the generated code using this customtype has a tag that we could use to check that this is a customtype and not just a struct with a UnmarshalJSON method.
https://github.com/gogo/protobuf/blob/master/test/thetest.pb.go#L1308

jdef:
Using gogo binaries from the latest master, I created a super simple .proto file and generated protos and ffjson for it. running `go test` on the package fails with:
```
$ go test ./tests/int64support/
--- FAIL: TestObjectJSON (0.00s)
        demopb_test.go:145: seed = 1500126240295096984, err = ffjson error: (*errors.errorString)cannot unmarshal tok:string into Go value for int64 offset=40 line=1 char=40
FAIL
FAIL    github.com/mesos/mesos-go/tests/int64support    0.002s
```

code: https://github.com/mesos/mesos-go/commit/9372b09c94a500d378cd834f448a6a3fab324b3f

awalterschulze:
Cool test :)

Could you maybe make that into a pull request, by changing that into a test that is a separate folder in the test folder?
Also I would like not to have a dependency on ffjson, but simple handwritten MarshalJSON and UnmarshalJSON methods.

Then you could also see if changing
https://github.com/gogo/protobuf/blob/master/jsonpb/jsonpb.go#L908
From:
```go
if target.CanAddr() {
```
To:
```go
if prop != nil && len(prop.CustomType) > 0 && target.CanAddr() {
```
Solves your problem?

jdef:
yes, that solves the problem. I started hacking on another PR. Rebuilding and running test case "regeneration" with the above change results in a ton of changes to gogo-generated code. Things lilke:

```
diff --git a/gogoproto/gogo.pb.go b/gogoproto/gogo.pb.go
index fa88040..1b59829 100644
--- a/gogoproto/gogo.pb.go
+++ b/gogoproto/gogo.pb.go
@@ -32,7 +32,7 @@ var E_GoprotoEnumPrefix = &proto.ExtensionDesc{
        ExtensionType: (*bool)(nil),
        Field:         62001,
        Name:          "gogoproto.goproto_enum_prefix",
-       Tag:           "varint,62001,opt,name=goproto_enum_prefix,json=goprotoEnumPrefix",
+       Tag:           "varint,62001,opt,name=goproto_enum_prefix",
        Filename:      "gogo.proto",
 }
 
@@ -41,7 +41,7 @@ var E_GoprotoEnumStringer = &proto.ExtensionDesc{
        ExtensionType: (*bool)(nil),
        Field:         62021,
        Name:          "gogoproto.goproto_enum_stringer",
-       Tag:           "varint,62021,opt,name=goproto_enum_stringer,json=goprotoEnumStringer",
+       Tag:           "varint,62021,opt,name=goproto_enum_stringer",
        Filename:      "gogo.proto",
 }
 
@@ -50,7 +50,7 @@ var E_EnumStringer = &proto.ExtensionDesc{
        ExtensionType: (*bool)(nil),
        Field:         62022,
        Name:          "gogoproto.enum_stringer",
-       Tag:           "varint,62022,opt,name=enum_stringer,json=enumStringer",
+       Tag:           "varint,62022,opt,name=enum_stringer",
        Filename:      "gogo.proto",
 }
 
```
expected?

awalterschulze:
I made the single line change and reran make and had no diffs except for that one line.

What protoc version are you using?

```
$ protoc --version
libprotoc 3.3.0
```

jdef:
Ah! I suspected that might be the problem, just ran out of time to test
earlier. Will bump and re-run

jdef:
that was part of it. the other part is related to that gzip problem, where
the generated descriptors differ between golang 1.7.x and 1.8.x because of
the inconsistent zero-val for the timestamp in the gzip header. nuking the
installed gogo binaries, updating PATH and bumping the golang version
resolved the crazy file changes I was seeing

awalterschulze:
Ah yes. This is why I only check the git diff on travis with the latest go
and protoc version. For example gofmt also changes sometimes with new go
versions.

jdef:
superseded by #308

PR #307, opened 2017-07-01
fix:305 only serialize protofields

### LINKED ISSUE #305: gogo jsonpb tries to serialize non proto field
We have a proto generated struct which also contains a non proto field for caching purposes, like so

```go
type PID struct {
	Address string `protobuf:"bytes,1,opt,name=Address,proto3" json:"Address,omitempty"`
	Id      string `protobuf:"bytes,2,opt,name=Id,proto3" json:"Id,omitempty"`

	p *Process //<-- this field
}
```

This works nice when serializing as a normal proto message.
But, when suing jsonpb, there is an error thrown that there is no encoder for `Process`.
Why is gogo jsonpb trying to serialize a field that is not part of the proto contract, has no proto tags and clearly is not a proto type?

rogeralsing:
This fixes #305
I am however not sure if this is the correct way to go about it, or if it even acceptable in terms of backwards compat to introduce this.

My understanding is however that when serializing using jsonpb, you only want to deal with fields defined in your proto contracts

### DIFF
--- a/jsonpb/jsonpb_test.go
+++ b/jsonpb/jsonpb_test.go
@@ -684,3 +684,33 @@ func (m *dynamicMessage) UnmarshalJSONPB(jum *Unmarshaler, json []byte) error {
 	m.rawJson = string(json)
 	return nil
 }
+
+// customFieldMessage implements protobuf.Message but is not a normal generated message type.
+type customFieldMessage struct {
+	someField string //this is not a proto field
+}
+
+func (m *customFieldMessage) Reset() {
+	m.someField = "hello"
+}
+
+func (m *customFieldMessage) String() string {
+	return m.someField
+}
+
+func (m *customFieldMessage) ProtoMessage() {
+}
+
+func TestMarshallCustomField(t *testing.T) {
+	rawJson := `{}`
+	m := &customFieldMessage{
+		someField: "hello world",
+	}
+	str, err := marshaler.MarshalToString(m)
+	if err != nil {
+		t.Errorf("an unexpected error occurred when marshalling JSONPBMarshaler: %v", err)
+	}
+	if str != rawJson {
+		t.Errorf("marshalling JSON produced incorrect output: got %s, wanted %s", str, rawJson)
+	}
+}
--- a/jsonpb/jsonpb_test.go
+++ b/jsonpb/jsonpb_test.go
@@ -426,7 +426,6 @@ var marshalingTests = []struct {
 	{"BoolValue", marshaler, &pb.KnownTypes{Bool: &types.BoolValue{Value: true}}, `{"bool":true}`},
 	{"StringValue", marshaler, &pb.KnownTypes{Str: &types.StringValue{Value: "plush"}}, `{"str":"plush"}`},
 	{"BytesValue", marshaler, &pb.KnownTypes{Bytes: &types.BytesValue{Value: []byte("wow")}}, `{"bytes":"d293"}`},
-	{"ignore non proto field", marshaler, &customFieldMessage{someField: "Ignore me"}, `{}`},
 }
 
 func TestMarshaling(t *testing.T) {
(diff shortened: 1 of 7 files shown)


awalterschulze:
I don't think this break backwards compatibility.
Everything without a protobuf tag was breaking anyway.
I would like a test though.

rogeralsing:
I'm on it

jsonpb/jsonpb_test.go
```diff
@@ -684,3 +684,33 @@ func (m *dynamicMessage) UnmarshalJSONPB(jum *Unmarshaler, json []byte) error {
 	m.rawJson = string(json)
 	return nil
 }
+
+// customFieldMessage implements protobuf.Message but is not a normal generated message type.
+type customFieldMessage struct {
```
rogeralsing: This is similar to the dynamicMessage but w/o the MarshalJSONPB code.
The message has no other fields than the custom field that should be ignored.
The accompanying test ensures that the  JSON output is `{}` of this type

rogeralsing:
Done.
There however seems to be a failing test now due to the change.
```
--- FAIL: TestMarshaling (0.00s)
        jsonpb_test.go:437: oneof, set: got [{}] want [{"title":"Grand Poobah"}]
```

It's not clear to me why this fails with this change. 
I will dig deeper

awalterschulze:
Good luck.

jsonpb/jsonpb_test.go
```diff
@@ -426,6 +426,7 @@ var marshalingTests = []struct {
 	{"BoolValue", marshaler, &pb.KnownTypes{Bool: &types.BoolValue{Value: true}}, `{"bool":true}`},
 	{"StringValue", marshaler, &pb.KnownTypes{Str: &types.StringValue{Value: "plush"}}, `{"str":"plush"}`},
 	{"BytesValue", marshaler, &pb.KnownTypes{Bytes: &types.BytesValue{Value: []byte("wow")}}, `{"bytes":"d293"}`},
+	{"ignore non proto field", marshaler, &customFieldMessage{someField: "Ignore me"}, `{}`},
```
rogeralsing: Custom field message added to marshal test data

rogeralsing:
OK, there seems to be code in the repo that relies on non mapped fields are getting serialized:
https://github.com/Serabe/protobuf/blob/a93a2631941938aa74d8c61e3603bca41ff93ee9/test/issue260/models.go#L15

My reasoning around jsonpb is that just as Protobuf binary format is cross platform, so is Proto JSON.
It's meant to mean the same even if you pass a Go message to a C# system etc.

Thoughts here?
Is this a no go?
Could an option be added to the marshaller to opt in for this?
cc @awalterschulze

awalterschulze:
I think its ok to go and add a protobuf tag for those two fields.
@Serabe what do you think?

rogeralsing:
All tests pass. 
I added my test data in the jsonpb_test.go, do you want this isolated as a separate folder under /tests instead?

awalterschulze:
I think I would just prefer the test in a different file, if you don't mind.
Some explicit like jsonpb_gogo_test.go
It just makes merging in future changes from golang/protobuf easier for me.

But the rest of the change looks perfect :)

awalterschulze:
You can also add yourself to the CONTRIBUTORS or AUTHORS as appropriate.

awalterschulze:
Good work :)
Thank you very much :)

Serabe:
Anything that requires reflection at some point requires struct tags. It is a known limitation on `typedecl`, but it can be manually added if needs be.

That said, if we add struct tags to all models, then we stop testing that `typedecl` works.

awalterschulze:
Hmmm thats a good point.

We should probably remove the tags again, but to get the jsonpb test to pass, maybe we should implement the JSONPBUnmarshaler interface for these structs?

That way we get best of both?

Serabe:
What would be needed to implement those?

awalterschulze:
I think this fixes the situation.
https://github.com/gogo/protobuf/commit/dda3e8acadcc9affc16faf33fbb229db78399245
What do you think?

Serabe:
👍

awalterschulze:
Ok great.
Thanks again for reviewing :)

PR #313, opened 2017-07-26
Fix unmarshal to correctly handle default value in maps

pattyshack:
According to the spec, map key/value entries should be deserialized the same way as a message with two 2 optional fields (https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility)

When the map entry's key (or value) is the default value for the specified type, the python/c++ implementation skips serializing that particular field.  The old Unmarshal implementation assumes the key is always in the serialized msg, which is not true.

For example,

proto definition:
```
syntax = "proto3";

message MapTest {
    map<string, string> str_str = 1;
}
```

python:
```
>>> msg =  MapTest() 
>>> msg.str_str[''] = ''
>>> print repr(msg.SerializeToString())
'\n\x00'
```
notice that neither key nor value are actually serialized (the map entry msg has msg len of zero)


awalterschulze:
Have you tested this against the non generated code?  Using proto.Marshal and proto.Unmarshal on a message without generated Marshal, Unmarshal and Size methods?
Because if it also fails then we should upgrade this discussion to golang/protobuf and keep the two implementations (generated and reflect) consistent.

pattyshack:
proto.Unmarshal is correctly handling "missing" default values.

proto.Marshal and the generated Marshal both unconditionally serializing the key/value fields, e.g., the above example is serialized to `[]byte{0xa, 0x4, 0xa, 0x0, 0x12, 0x0}`.  This is correct, but suboptimal (we can leave Marshal alone for consistency)

awalterschulze:
I'll look more in depth into this asap, this week is just a little crazy.

awalterschulze:
The checks are failing.
Which version of protoc are you using.  The diff checker expects you to use protoc version 3.3 and go version 1.8, but I suspect this is only a protoc issue.

awalterschulze:
Also could you add a test, that failed before this change?  To make sure we don't regress back to the current state.

awalterschulze:
I see you also added some really nice code cleanup while making this improvement, thanks :)

pattyshack:
hmm, i may have screwed up rebasing.  let me drop this pull request and redo things ...

PR #316, opened 2017-07-31
grouping similar imports in .pb.go file

wishmaster-dk:
I suggest to group similar imports in .pb.go file instead of making duplicates. 

Example. At the moment these imports in .proto file produce duplicated imports in resulting .pb.go file.
**.proto file**
`import "google/protobuf/wrappers.proto";`
`import "google/protobuf/timestamp.proto";`
**.pb.go file**
`import google_protobuf "google/protobuf"`
`import google_protobuf1 "google/protobuf"`

Nice feature, but this causes issues with parsing .pb.go file (we use code generation based on .pb.go file). 

This tiny PR removes duplication of imports and leaves only one import instead of a few similar.
**.pb.go file**
`import google_protobuf "google/protobuf"`

### DIFF
--- a/protoc-gen-gogo/generator/generator.go
+++ b/protoc-gen-gogo/generator/generator.go
@@ -718,23 +718,14 @@ func RegisterUniquePackageName(pkg string, f *FileDescriptor) string {
 	// Convert dots to underscores before finding a unique alias.
 	pkg = strings.Map(badToUnderscore, pkg)
 
-	var i = -1
 	var ptr *FileDescriptor = nil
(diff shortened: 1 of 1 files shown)


awalterschulze:
Could you try regenerating with protoc version 3.3, so that the checks pass.

I would then want to ask some people to test this on their projects, since imports are the one thing that is a little under tested in the gogoprotobuf project.

I really hope this works.

awalterschulze:
We think this has been fixed by https://github.com/gogo/protobuf/pull/341

PR #318, opened 2017-07-31
Fix unmarshal to correctly handle default value in maps

pattyshack:
original context in #313


pattyshack:
This address both the zipped descriptor issue (go1.8 updated the zip implementation) and testing.  Let me know if there's anything else.

plugin/unmarshal/unmarshal.go
```diff
@@ -747,56 +789,50 @@ func (p *unmarshal) field(file *generator.FileDescriptor, msg *generator.Descrip
...
+
+			p.P(`var wire uint64`)
+			p.decodeVarint("wire", "uint64")
+			p.P(`fieldNum := int32(wire >> 3)`)
+
+			p.P(`if fieldNum == 1 {`)
+			p.In()
+			p.mapField("mapkey", false, m.KeyAliasField)
+			p.Out()
+			p.P(`} else {`)
```
awalterschulze: else if fieldNum == 2
else // hmmm.  I guess ignore, since we have to stay compatible with a map object which was manually defined and had extra fields.

test/mapdefaults/Makefile
```diff
@@ -0,0 +1,32 @@
...
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+regenerate:
+	go install github.com/gogo/protobuf/protoc-min-version
+	go install github.com/gogo/protobuf/protoc-gen-gogofast
+	protoc-min-version --version="3.0.0" --proto_path=../../../../../:../../protobuf/:. --gogofast_out=. map.proto
```
awalterschulze: I think we should rather use protoc-min-combo, since I would really like to see how all implementations are compatible with each other.

awalterschulze:
Sorry for the delayed reply.

plugin/unmarshal/unmarshal.go
```diff
@@ -747,56 +793,50 @@ func (p *unmarshal) field(file *generator.FileDescriptor, msg *generator.Descrip
...
+
+			p.P(`if fieldNum == 1 {`)
+			p.In()
+			p.mapField("mapkey", false, m.KeyAliasField)
+			p.Out()
+			p.P(`} else if fieldNum == 2 {`)
+			p.In()
+			p.mapField("mapvalue", gogoproto.IsCustomType(field), m.ValueAliasField)
+			p.Out()
+			p.P(`}`)
```
awalterschulze: Are we skipping the field if the message has a field 3?
We still need to process the bytes and move on to the next field, otherwise we are not properly ignoring it.

test/mapdefaults/Makefile
```diff
@@ -0,0 +1,33 @@
+# Protocol Buffers for Go with Gadgets
+#
+# Copyright (c) 2015, The GoGo Authors. All rights reserved.
```
awalterschulze: 2017

test/mapdefaults/Makefile
```diff
@@ -0,0 +1,33 @@
...
+# LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+# DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+# THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+regenerate:
+	go install github.com/gogo/protobuf/protoc-gen-combo
+	go install github.com/gogo/protobuf/protoc-gen-gogofast
+	protoc-gen-combo --version="3.0.0" --proto_path=../../../../../:../../protobuf/:. --gogofast_out=. map.proto
```
awalterschulze: I think it should be gogo_out, since we don't want it to automatically generate the marshaler and unmarshaler.  That is what the extensions are for.

test/mapdefaults/combos/both/map_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/gogo/protobuf/proto"
+)
+
+func TestUnmarshalImplicitDefaultKeyValue1(t *testing.T) {
+	// The follow message is generated via python:
+	//
+	// map = MapTest()
+	// msg.str_str['foo'] = ''
+	// msg.str_str[''] = 'bar'
```
awalterschulze: Maybe we could generate this with here with a proto go message.

```proto
message Map {
   repeated OldMap OldMap = 1;
}

message OldMap {
   optional string key = 1;
   optional string value = 2;
}
```
```go
m := &Map{}
m.OldMap = append(m.OldMap, &OldMap{Key: "foo", Value: ""}
m.OldMap = append(m.OldMap, &OldMap{Key: "", Value: "bar"}
...
```

This way we can also test whether we are skipping field 3, by adding a field number 3 to OldMap.

plugin/unmarshal/unmarshal.go
pattyshack: will return an error if there are additional fields.

awalterschulze: It shouldn't return an error, because then if breaks compatibility.
The unknown fields should simply be skipped.

awalterschulze: This might help
https://github.com/gogo/protobuf/pull/318/files#diff-e0188cb7643b9790acd32abc395a5659L1275

pattyshack: Sure. I can add code to skip bad tags.  but note that the default (non-codegen) implementation does not handle this case either https://github.com/gogo/protobuf/blob/master/proto/decode.go#L877

awalterschulze:
Good point about the fact that the reflect version does not support unknown fields
https://github.com/gogo/protobuf/blob/master/proto/decode.go#L877
It seems that golang/protobuf is thinking about whether they should skip it or not, given the TODO right about it.
I think they will add a skip in the future.
I opened an issue https://github.com/golang/protobuf/issues/403

awalterschulze:
I think we can merge this in the mean time.  If we discover that this really should be an error, then we can make a fix again.

awalterschulze:
Thanks for some great work here :)

PR #321, opened 2017-08-11
Only rename size fields if sizer is enabled.

devnev:
The Size method is only created if the sizer plugin is enabled, and other parts of the codebase seem to handle this case. The helpers here were assuming that either Size or ProtoSize exists, which is not true, especially if the base message was generated with the google golang plugin.

### DIFF
--- a/protoc-gen-gogo/generator/generator.go
+++ b/protoc-gen-gogo/generator/generator.go
@@ -2049,8 +2049,11 @@ func (g *Generator) generateMessage(message *Descriptor) {
 	for _, n := range methodNames {
 		usedNames[n] = true
 	}
-	if !gogoproto.IsProtoSizer(message.file, message.DescriptorProto) {
+	if gogoproto.IsSizer(message.file, message.DescriptorProto) {
 		usedNames["Size"] = true
+	}
+	if gogoproto.IsProtoSizer(message.file, message.DescriptorProto) {
+		usedNames["ProtoSize"] = true
 	}
 	fieldNames := make(map[*descriptor.FieldDescriptorProto]string)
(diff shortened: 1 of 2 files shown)


awalterschulze:
Could you add a test that shows how it broke before this change?
So that we don't regress in future.

awalterschulze:
hello?

schleppy:
I would like to volunteer my help in getting this merged.  This is a necessary change for our continued use of the library.  What sort of test are you looking for?  Is there an existing example exercising another feature that might work as a basis?  Essentially what the test(s) would do is ensure there is no `Size` and `GetSize()` created when `sizer` is not used.

jmarais:
Hi, Thanks for volunteering.
I didn't look at this issue initially, but correct me if I am wrong. I think the issue is we assume we are generating a Size() method, even without enabling the gogo-option. 
This means we will always substitute fields ```Size -> Size_```.

>  Is there an existing example exercising another feature that might work as a basis? What sort of test are you looking for? 

Here an example of tests involving ```Size``` and ```ProtoSize```.

[protosize](https://github.com/gogo/protobuf/tree/master/test/protosize)

You could create a similar tests and assert the resulting structure has the expected methods/fields with the combination of the ```sizer_all```,```protosizer_all```, no size options?
That would assert that the generation of the structs are as expected with the relevant size options?


> Essentially what the test(s) would do is ensure there is no Size and GetSize() created when sizer is not used.
Yeh

vtolstov:
does it possible to not modify struct member Size to Size_ ?i need to expose this member as-is with name Size.

icefed:
Anyone can fix it? I need this too.

PR #328, opened 2017-08-26
proto: use sync.Pool for temporary Buffer types

irfansharif:
Lower GC pressure for whenever temporary `Buffer` types are allocated, evidently it's an issue [grpc-go](https://github.com/grpc/grpc-go) ran into in the past. Additionally remove TODOs left regarding exactly this.

+cc https://github.com/grpc/grpc-go/pull/1478#issuecomment-325058960, https://github.com/golang/protobuf/pull/418

### DIFF
--- a/proto/decode.go
+++ b/proto/decode.go
@@ -409,7 +409,17 @@ func UnmarshalMerge(buf []byte, pb Message) error {
 	if u, ok := pb.(Unmarshaler); ok {
 		return u.Unmarshal(buf)
 	}
-	return NewBuffer(buf).Unmarshal(pb)
+
+	b := bufPool.Get().(*Buffer)
+	b.Reset()
+	b.SetBuf(buf)
+	err := b.Unmarshal(pb)
+	// We swap out the internal buffer with a nil slice so that the user
+	// provided slice is not mutated subsequently.
+	b.SetBuf(nil)
+	bufPool.Put(b)
+
+	return err
 }
 
 // DecodeMessage reads a count-delimited message from the Buffer.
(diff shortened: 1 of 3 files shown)


awalterschulze:
Lets see what happens on golang protobuf first, since gogo is just a fork.

irfansharif:
Fair, will ping here if hear back on that front or in a week, whatever comes first.

irfansharif:
https://github.com/golang/protobuf/pull/418#issuecomment-325467576 'Twas rejected. Still worthwhile pulling in here (I think).

awalterschulze:
I am going to have to pull in their changes. It is going to be extremely
painful already. And most gogoprotobuf users prefer the generated version
of marshal and unmarshal. How about we wait to see their changes?

irfansharif:
> most gogoprotobuf users prefer the generated version of marshal and unmarshal

That change would actually most strongly benefit this group of users. (We're one too, cockroachdb/cockroach, which was the original motivation here given the measurable impact.)
In any case, like I mentioned there, the ~50 line diff only changes calls to `NewBuffer` in the right hotspots. I should think this would hardly be painful given what seems to be a bigger set of changes across the board (and not sure when it'll be exported).

I do see you point however, perfectly fine either way.

awalterschulze:
If you generated code then you would exit before getting to your changes

```go
if u, ok := pb.(Unmarshaler); ok {
  		return u.Unmarshal(buf)
}
```

```go
if m, ok := pb.(Marshaler); ok {
  		return m.Marshal()
}
```

Or am I missing something?

irfansharif:
welp, you had it right the first time, this would affect only those without `{Unm,M}arshal` methods. I was conflating this with grpc/grpc-go's usage pattern in using proto.Buffers irregardless of `{Unm,M}arshaler` satisfying messages. My apologies, thanks for pointing that out. 

> most gogoprotobuf users prefer the generated version of marshal and unmarshal

Makes this slightly moot here considering, but I have another solution for  grpc/grpc-go#1478 so thanks anyhow! Feel free to close this if necessary.

awalterschulze:
I would like this to work for grpc
https://github.com/gogo/protobuf/blob/master/codec/codec.go

But then they must create a codec for each concurrent connection and try to reuse them.
So grpc can handle the sync.Pool of codecs and then codecs can be stupid.
But thats my idea, which is not flying, but I can't seem to find the reference discussion on this.

awalterschulze:
No problem btw.
I wish you luck, the grpc handling on protos is super inefficient and really needs to be fixed.

PR #336, opened 2017-09-22
Visit field types of messages even if a type should not be generated

erizocosmico:
If you use a message type imported from another package in a message with `gogoproto.typedecl = false` the import of that package (if nothing else uses it) will be aliased to `_` instead of the actual name, leading to a compilation error.

This iterates over all the message fields even if we don't have to generate a type for it only to visit the types and mark the imports as used if needed.


awalterschulze:
Thats great.  Could you include a test, to make sure we don't accidentally remove this fix.

erizocosmico:
Sure! Done

PR #339, opened 2017-10-03
[WIP] LittleEndian stuff

diamondo25:
- Fix benchmarking
- Fix error logs when compile error happens (cut out only the lines of the error).
- Write compiled file to a temporary file, so you can check it out manually


awalterschulze:
Sorry for not using this, but you have helped more than you can imagine.
Thanks again for reporting and putting in all this effort.
 I really appreciate it.

Issue #338, opened 2017-10-03
Can I generate imports without the import alias (or without the long underscored import path)?


meling:
In our plugin, we have until now used a custom `GenerateImports(file *generator.FileDescriptor)` method, but it turned out to be unreliable in certain scenarios (see (*) below). So we wanted to switch to using the `generator.NewPluginImports(g.Generator)` thing. However, right now, I'm getting 
```go
import golang_org_x_net_trace "golang.org/x/net/trace"
import google_golang_org_grpc_codes "google.golang.org/grpc/codes"
import google_golang_org_grpc_status "google.golang.org/grpc/status"
```
which doesn't work for us since our plugin is generated from template code, which can't use the `Single.Use()` method. It would require us to hardcode these underscored import paths into our templates, which would be quite tedious, error prone, and make our templates quite unreadable.

My question is: is there a way to generate these imports without the underscored import alias??

(PS: I don't quite understand the need for this import alias thing, but I'm sure there is an explanation. But is it always necessary?)

(*) There are some more details concerning this [issue here](https://github.com/relab/gorums/issues/38), but hopefully, the above description should be clear. Here is an example [template](https://github.com/relab/gorums/blob/master/dev/calltype_common_definitions.tmpl) being used by our plugin.

awalterschulze:
As far as I remember the long alias names in gogoprotobuf's generated code is simply a way to avoid most of the package name collisions and try to guarantee some uniqueness. You are welcome to try and change the import logic in gogoprotobuf, but its not fun and I would want a few production projects to test the change, since its quite convoluted.

Here are a couple of work arounds:

1) You can build Single.Use() into the template object as separate methods:

  - ImportContext
  - ImportCodecs

2) Or you could add a function or method to the object that takes a string:

  - {{import "context"}}.Context
  - {{import  "golang.org/x/net/trace"}}

3) Parse your generated go code for imports and add them to the ast, before outputting the generated code.

Sorry.  I know this is not the answer you were hoping for.

meling:
I've tried to make a few small changes to the import logic that solves my problem. I've uploaded a PR with my fix, but clearly it would require much more scrutiny.

meling:
[...]

PR #341, opened 2017-10-05
Fix to avoid printing import aliases when not needed

meling:
It required a minor change in the `grpc` package, but otherwise the tests passed (surprisingly). This PR solves my own use case in issue #338.

PS: I realize this may potentially be an invasive change, and so should be tested widely on use cases outside the main repo.


protoc-gen-gogo/generator/generator.go
```diff
@@ -1176,6 +1181,9 @@ func (g *Generator) addInitf(stmt string, a ...interface{}) {
 
 func (g *Generator) PrintImport(alias, pkg string) {
 	statement := "import " + alias + " " + strconv.Quote(pkg)
+	if alias == pkg[strings.LastIndex(pkg, "/")+1:] {
```
awalterschulze: I suspect this won't always work, since the packageName can differ from the packagePath.

protoc-gen-gogo/generator/generator.go
```diff
@@ -715,8 +715,13 @@ var pkgNamesInUse = make(map[string][]*FileDescriptor)
 // Pkg is the candidate name.  If f is nil, it's a builtin package like "proto" and
 // has no file descriptor.
 func RegisterUniquePackageName(pkg string, f *FileDescriptor) string {
-	// Convert dots to underscores before finding a unique alias.
-	pkg = strings.Map(badToUnderscore, pkg)
+	// Only make pkg names unique if they are non-standard lib
+	if f != nil {
```
awalterschulze: Nice trick here :)

awalterschulze:
I also wonder about a case where someone is importing github.com/gogo/protobuf/proto and github.com/golang/protobuf/proto
Which is what happens when someone uses the `goproto_registration` plugin.

awalterschulze:
When we have thought about these cases, then hopefully we are at a place where we can ask a gogoprotobuf user to test this one their system.

meling:
I ran the tests in `github.com/gogo/protobuf/test/registration` locally, and the following import statements are produced:
```go
import proto "github.com/gogo/protobuf/proto"
import golang_proto "github.com/golang/protobuf/proto"
```
Of course, there could be other issues that I've overlooked. Let me know if there is any specific tests beyond this I should run. 

I've also checked a few other places and it seems that some duplicate imports of the same package under different aliases are eliminated, e.g. in `vanity/test/faster/vanity.pb.go`. 

About the Travis CI build failing, I assume it is caused by some diff check of the current output against previous output. I was unable to find a specific test failure or error message in the log. (Perhaps if I committed also the regenerated `.pb.go` files etc. this problem would go away?)

(I'm sorry if my unfamiliarity with the dev process is causing this. Please advise on how to do things properly, if I'm doing it wrong; maybe there is a contributor guide that I should read?)

awalterschulze:
Those proto imports are beautiful :)

On Travis you should be able to click on details and see the printout.
But it is as you suspect, it is simply a diff that is failing, that wants the regenerated code to be commited.  Simply calling `make` in the project root should fix this.

Your doing perfectly great :)
But adding a contributing guidelines doc is maybe not a bad idea.  I just haven't found one in the wild which I thought was a good read yet, but I am open to suggestions and would like to know where you are struggling.
From my end though, you are doing great.

I think this is ready for a first outside tester.

After that I think it would be great to add a little test to gogoprotobuf, where multiple imports end in the same suffix, just like `proto`, but without requiring more dependencies.

awalterschulze:
@tamird , can I ask you a favor.  Can you test this branch on your codebase?

protoc-gen-gogo/grpc/grpc.go
```diff
@@ -129,11 +129,15 @@ func (g *grpc) GenerateImports(file *generator.FileDescriptor) {
 	if len(file.FileDescriptorProto.Service) == 0 {
 		return
 	}
-	g.P("import (")
-	g.P(contextPkg, " ", strconv.Quote(path.Join(g.gen.ImportPrefix, contextPkgPath)))
-	g.P(grpcPkg, " ", strconv.Quote(path.Join(g.gen.ImportPrefix, grpcPkgPath)))
-	g.P(")")
-	g.P()
+	grpcImports := []string{
+		path.Join(g.gen.ImportPrefix, contextPkgPath),
```
tamird: this bit is incorrect - there is no need for this `path.Join` - the `imports` object already takes care of the import prefix. In CockroachDB, this causes the import prefix to appear twice.

protoc-gen-gogo/generator/generator.go
tamird: This comment is still wrong (this code only makes the names unique if they are *not* builtin).

However, this makes me think that this comment is just not useful. Perhaps instead of explaining what this code is doing (which is pretty clear just from reading the code and the method comment), explain **why** it is doing it?

meling: How about this:
```go
	// Ensure that package names across multiple imports are unique, unless they
	// are builtin or standard lib packages, in which case they should not conflict.
```

awalterschulze:
I meant to put this comment in the main thread

I wonder if a package like sortkeys is a f!=nil package and if so I wonder what happens when I also try to import a proto file with package name sortkeys.

Maybe it is time to start adding a test like this one?

meling:
Yeah, I guess they should be squashed into a single commit at some point when it is ready to be merged, but I'm not sure how legible the commits will be anyway with all the generated noise. Anyway, I planning to take a stab at adding the tests that @awalterschulze requested; these last few commits are just to bring my fork up to dated with the master branch.

meling:
@awalterschulze I've prepared a new test to check that my proposed patch still works when a proto files uses a generated package (from another proto file) that conflicts with the name of an internal package (in this case, `sortkeys` as suggested earlier). Let me know if you want something more in addition to this.

meling:
I discovered a problem, if the proto package name is `proto`, and there is also an import like this:
```go
import proto "github.com/gogo/protobuf/proto"
```
My fix won't detect this. I'm thinking about a fix.

awalterschulze:
I am really positive about this change.
I can squash the commits when I merge.
Also github makes it easy to review the generated code. I don't see it as
noise.
Good luck with the final issue

meling:
I've tried a wide range of things to make the imports more readable, including trying to incorporate #316, but that fails with my new tests in `test/importduplicate`. Hence, unless you think my tests are unreasonable (importing from a sub package `proto/proto.proto`), I think you can close #316. 

**Anyway, I've resolved the issues I discovered yesterday, and so this should be ready to be merged.**

PS: If you import a type whose name overlap with the package name of the proto file, then that type gets named `type1`. For example, in test `package types`, there will be an import like this:
```go
import types1 "github.com/gogo/protobuf/types"
```
This is strictly not necessary, as Go allows importing `types` even though the package is also `types`. However, fixing that issue is beyond the need for my other project, and it does not appear trivial to fix.

awalterschulze:
I think thats fair.
Really great work :)
Thank you very much.  I really appreciate it.

tamird:
Womp womp, no squash

awalterschulze:
My bad.

PR #352, opened 2017-11-09
Test with latest Go patch versions

AlekSi:


awalterschulze:
If we do this we also need to update the Readme.md which currently mentions specific go versions.

AlekSi:
PTAL

awalterschulze:
Awesome :)

Please change 3.4.1 to 3.4.0 since 3.4.1 does not have a proper download link and rebuilding protoc from source is very slow, as you can see with protoc 2.6.1

AlekSi:
PTAL :)

awalterschulze:
Nice thank you :)

PR #355, opened 2017-11-17
Mentioning the bit about proto_path up front

agnivade:
I had to spend quite some time trying to debug and understand how to setup the proto_path. The example in the Makefile just shows some relative paths which are not very helpful.

I think mentioning them directly in the README will save every one a lot of time. (It sure would have done for mine !)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -95,14 +95,16 @@ Other binaries are also included:
 Installing any of these binaries is easy.  Simply run:
 
     go get github.com/gogo/protobuf/proto
-    go get github.com/gogo/protobuf/protoc-gen-gogo{fast|faster|slick}
+    go get github.com/gogo/protobuf/{binary}
     go get github.com/gogo/protobuf/gogoproto
 
-These binaries allow you to use gogoprotobuf [extensions](https://github.com/gogo/protobuf/blob/master/extensions.md).
+These binaries allow you to use gogoprotobuf [extensions](https://github.com/gogo/protobuf/blob/master/extensions.md). You can also use your own binary.
 
 To generate the code, you also need to set the include path properly.
 
-    protoc -I=. -I=$GOPATH/src -I=$GOPATH/src/github.com/gogo/protobuf/protobuf --gogo{fast|faster|slick}_out=. myproto.proto
+    protoc -I=. -I=$GOPATH/src -I=$GOPATH/src/github.com/gogo/protobuf/protobuf --{binary}_out=. myproto.proto
+
+Note that in the protoc command, {binary} does not contain the initial prefix of "protoc-gen". 
 
 ### Most Speed and most customization
 
--- a/Readme.md
+++ b/Readme.md
@@ -95,10 +95,14 @@ Other binaries are also included:
 Installing any of these binaries is easy.  Simply run:
 
     go get github.com/gogo/protobuf/proto
-    go get github.com/gogo/protobuf/{binary}
+    go get github.com/gogo/protobuf/protoc-gen-gogo{fast|faster|slick}
     go get github.com/gogo/protobuf/gogoproto
 
 These binaries allow you to use gogoprotobuf [extensions](https://github.com/gogo/protobuf/blob/master/extensions.md).
+
+To generate the code, you also need to set the include path properly.
+
+    protoc -I=. -I=$GOPATH/src -I=$GOPATH/src/github.com/gogo/protobuf/protobuf --gogo{fast|faster|slick}_out=. myproto.proto
 
 ### Most Speed and most customization
 


awalterschulze:
I am not sure about the part in square brackets. Maybe leave that and
simply explain the gogo case. Then below you can explain that you can also
use other binaries or even your own custom one.

awalterschulze:
But this is definitely a step in the right direction. Thank you for the
effort.

awalterschulze:
Sorry about the vacation time.  That got in the way here.

I think maybe skewing the example to gogo_out and protoc-gen-gogo, so that its immediately useful and then adding a sentence below about that how this could be replaced with your own binary.

What do you think?

agnivade:
Alright. But gogo_out is the same as usual protobuf right ? How about using gogofaster_out ?

awalterschulze:
Ok wait.  My bad.

awalterschulze:
Now that I read it again, I think what you have here is perfect.
Sorry I forgot in which section we were :)

awalterschulze:
Thank you for the clarification in the readme.
I think this will help a lot of newcomers :)

awalterschulze:
Btw @agnivade great blog post, but I guess I am a bit biased :)
Would you like to add it to the Readme.md in the mentioned section?

Could I maybe ask a favour, that you clarify this statement:
```
Yes, using gogoproto, you cannot use protocol buffer 3 because it does not support extensions. There is an active issue open which discusses this in further detail.
```
By saying
```
Yes, using gogoproto, you cannot use proto3 if you intend to share your protobuf definition with languages which do not support proto2, like php, because proto3 does not support extensions. There is an active issue open which discusses this in further detail.
```
This means that mostly you can use proto3 with gogoprotobuf.
Also another way to get around the issue would be to not use any extensions and simply call gogoslick_out, for the case in the blog post.

agnivade:
Thanks for reading the blog post :) 

Hmm .. interesting. I actually got a compiler error when I tried to use `syntax = "proto3";`. Is it because I used gogofaster and not gogoslick ?

awalterschulze:
Gogo_out is the same as go_out except that it changes the imports to import
gogoprotobuf instead of golang protobuf and it also takes all your
extensions into account when generating code. Gogofaster, gogoslick, etc
allows you to specify less extensions.

Could you please share the compile error and how to reproduce it. That
sounds like a possible bug.

awalterschulze:
Thank you for writing a blog post. I really appreciate it.

agnivade:
I just updated the blog post. Will send a PR soon.

Regarding the error. This is my proto file -
```protobuf
syntax = "proto3";
package xxxx;

option (gogoproto.gostring_all) = true;
option (gogoproto.goproto_stringer_all) = false;
option (gogoproto.stringer_all) =  true;
option (gogoproto.marshaler_all) = true;
option (gogoproto.unmarshaler_all) = true;

// For tests
option (gogoproto.equal_all) = true;

// This is the message sent to the cloud server
message ClientMessage {
  string Field1 = 1 [(gogoproto.nullable) = false];
  string Field2 = 2 [(gogoproto.nullable) = false];
  int64 Timestamp = 3 [(gogoproto.nullable) = false];
}
```

And this is the error -
```
protoc -I=. -I=/home/agniva/play/go/src -I=/home/agniva/play/go/src/github.com/gogo/protobuf/protobuf --gogofaster_out=. *.proto
ERROR: field ClientMessage.Field1 is a native type and in proto3 syntax with nullable=false there exists conflicting implementations when encoding zero values--gogofaster_out: protoc-gen-gogofaster: Plugin failed with status code 1.
```

This works fine - 
```protobuf
syntax = "proto2";
package xxxx;

option (gogoproto.gostring_all) = true;
option (gogoproto.goproto_stringer_all) = false;
option (gogoproto.stringer_all) =  true;
option (gogoproto.marshaler_all) = true;
option (gogoproto.unmarshaler_all) = true;

// For tests
option (gogoproto.equal_all) = true;

// This is the message sent to the cloud server
message ClientMessage {
  required  string Field1 = 1 [(gogoproto.nullable) = false];
  required  string Field2 = 2 [(gogoproto.nullable) = false];
  required  int64 Timestamp = 3 [(gogoproto.nullable) = false];
}
```

awalterschulze:
Ah so the error message says it all, maybe not clearly enough though :( In
proto3 you don't need nullable=false for native types like string, since
proto3 already does not generate pointers. If you want to update the error
message to make it more clear that would be great.

If you remove nullable=false then it should work. Could you try and let me
know?

agnivade:
Confirmed. Works. :smile: 

Yes, the error message could certainly improve. Especially the last part - `there exists conflicting implementations when encoding zero values--gogofaster_out: protoc-gen-gogofaster`. The user doesn't know exactly what needs to be done.

awalterschulze:
Could you please help, since you are so much better at explaining and an
actual user and make a pull request.

Or a suggestion for a better error message?

agnivade:
Absolutely.

PR #357, opened 2017-11-30
plugin/populate: avoid loops on non top-level packages

### LINKED ISSUE #207: Dependency of dependency causing WARNINGS at compile time
When using protoc-gen-gogo I see messages like this for depenencies of a dependency:

```
2016/10/03 12:13:29 protoc-gen-gogo: WARNING: failed finding publicly imported dependency for .<package_message>, used in <my_package>/<my_package>.proto
```

When explicitly importing the package in the affected protofile, this warning disappears, but obviously I then instead get the warning that there are unused imports. This seems to imply that there is a problem with the dependency-of-dependency resolution.

benesch:
The previous scheme for detecting when a type belonged to a different
package only considered the topmost component of the type name, when it
in fact needs to consider all but the last component of the type name.

Fix #207 (again).

### DIFF
--- a/plugin/populate/populate.go
+++ b/plugin/populate/populate.go
@@ -443,7 +443,7 @@ func (p *plugin) GenerateField(file *generator.FileDescriptor, message *generato
 	}
 }
 
-func (p *plugin) hasLoop(field *descriptor.FieldDescriptorProto, visited []*generator.Descriptor, excludes []*generator.Descriptor) *generator.Descriptor {
+func (p *plugin) hasLoop(pkg string, field *descriptor.FieldDescriptorProto, visited []*generator.Descriptor, excludes []*generator.Descriptor) *generator.Descriptor {
 	if field.IsMessage() || p.IsGroup(field) || p.IsMap(field) {
 		var fieldMessage *generator.Descriptor
 		if p.IsMap(field) {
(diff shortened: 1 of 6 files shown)


tschottdorf:
💛

awalterschulze:
Beautiful :)

Could you just regenerate the code, because the diff check is failing.
I usually do this by simply running `make` in the root of the project.

benesch:
Heh, turns out that one-line diff was actually pointing out that my new approach was (still) wrong. Type names can include the names of other types, so simply stripping off the last component isn't sufficient to determine the package either.

Updated with a new approach which plumbs the package name from the file descriptor. I'm hopeful that's bulletproof.

awalterschulze:
Is it possible to also include your example as a test?

benesch:
Indeed, the ref I just pushed does exactly that :)

awalterschulze:
Awesome I'll be at a computer soon.

benesch:
Whew, it's green! Thanks for the crazy fast response time, @awalterschulze.

awalterschulze:
Perfect :)
Thank you very much.

awalterschulze:
Any time :) 
Well ... not any time unfortunately :(
, but I try my best.

PR #358, opened 2017-11-30
README.md: add note about gogoproto extensions include path

benesch:


awalterschulze:
This is only true if you `import "gogoproto/gogo.proto";` but I usually import `import "github.com/gogo/protobuf/gogoproto/gogo.proto";` and then adding $GOPATH/src is enough.
See example
https://github.com/gogo/protobuf/tree/master/test/example

benesch:
Ah, fair enough. I suppose we're peculiar over at CockroachDB in that we don't like to include the whole GOPATH lest we pick up some non-vendored protos. Closing!

Issue #360, opened 2017-12-02
linter complains on code in generated Equal implementation


temoto:
First, huge thanks for improved protobuf.

Code generated by gogo/slick:
```
func (this *Request) Equal(that interface{}) bool {
	if that == nil {
		if this == nil {
			return true
		}
		return false
	}
```

`gosimple` linter complain:
> /path/bla/bla.pb.go:349:3:warning: should use 'return <expr>' instead of 'if <expr> { return <bool> }; return <bool>' (gosimple)

Generated code could be reduced down to:
```
func (this *Request) Equal(that interface{}) bool {
	if that == nil {
		return this == nil
	}
```

Thanks for gogo/protobuf. Have a great time.

awalterschulze:
My pleasure :)

Is this the only remark the linter had?

temoto:
Also it doesn't like repeated string literals in `*valueToString()` and `GoString()`. I guess linter wants `"nil"` hoisted into constant declaration.

awalterschulze:
I would be willing to review the code if you are willing to make a pull request to fix these, if possible?

temoto:
* Fixing `return this == nil` was super easy, nice code.
* Seems I'm not smart enough to fix `"nil"` constant. Naive approach would create duplicate declaration in sibling `second.pb.go` in same package.
* Irrelevant thought on `p.In()/p.Out()` - wouldn't it be better to produce "raw" code without indentation and fix everything with `gofmt` run?

awalterschulze:
On p.In and p.Out.
Maybe yes.
In my other code generating project [goderive](https://github.com/awalterschulze/goderive) I generate gofmted code the first time, using p.In and p.Out.  I find that running gofmt over a few 100 000 lines of code becomes quite slow and tedious.
But if you code generators output go fmted code and your editor gofmts your code when you save, then you technically don't have to ever run gofmt over your whole code base.
So in the interest of a fast compile (and code generation) feedback loop, I think we actually need to put in more of an effort in gogoprotobuf to generate gofmted code and that way we can speed up some developers' feedback loop.

temoto:
Understood. Have a great time.

awalterschulze:
I hope it makes sense.

temoto:
#perfmatters yo 👍

shurcooL:
[...]

PR #361, opened 2017-12-03
plugin/equal: "return this == nil" to satisfy gosimple linter

temoto:
https://github.com/gogo/protobuf/issues/360


awalterschulze:
Thank you very much :)

PR #363, opened 2017-12-20
Update to protobuf 3.5.1 and minor cleanups for Golang 1.10

peter-edge:
This does the following:

- Update the latest version of Protobuf from 3.4.0 to 3.5.0.1
- Remove some trailing newlines in install-protobuf.sh
- Add a dummy non-test file for `go vet` to work in Golang 1.10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
 script:
   - PATH=/home/travis/bin:$PATH make buildserverall
   - echo $TRAVIS_GO_VERSION
-  - if [[ "$PROTOBUF_VERSION" == "3.5.0" ]] && [[ "$TRAVIS_GO_VERSION" =~ ^1\.9\.[0-9]+$ ]]; then ! git status --porcelain | read || (git status; git diff; exit 1); fi
+  - if [[ "$PROTOBUF_VERSION" == "3.5.1" ]] && [[ "$TRAVIS_GO_VERSION" =~ ^1\.9\.[0-9]+$ ]]; then ! git status --porcelain | read || (git status; git diff; exit 1); fi
 
 language: go
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,7 @@ before_install:
 
 script:
   - PATH=/home/travis/bin:$PATH make buildserverall
-  - echo $TRAVIS_GO_VERSION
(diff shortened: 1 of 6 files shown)


awalterschulze:
Only new protobuf versions that are available for download with wget are supported.  Maybe choose a less minor version?

peter-edge:
Whoops, fixed

awalterschulze:
Seems `CHECKDIFF` should only be `1` if go is 1.10 (gofmt) and protoc is 3.5.0

peter-edge:
Updated to only run on Golang 1.9

awalterschulze:
Ah damn I forgot we don't have the go version in the matrix vars.  So now the CHECK_DIFF var does not make sense anymore

I suggest this and removing CHECK_DIFF entirely
```
if [[ "$PROTOBUF_VERSION" == "3.5.0" ]] && [[ "$TRAVIS_GO_VERSION" == "1.10" ]]; then ! git status --porcelain | read || (git status; git diff; exit 1); fi
```

Also please keep `echo $TRAVIS_GO_VERSION`

And then we need to add go 1.10 to the go versions being tested, since otherwise we are not checking the diff.

Then we also need to add these new versions that we are testing to the Readme.md

peter-edge:
I think this is ready to go.

.travis.yml
```diff
@@ -10,10 +10,11 @@ before_install:
 script:
   - PATH=/home/travis/bin:$PATH make buildserverall
   - echo $TRAVIS_GO_VERSION
-  - if [ "$CHECK_DIFF" == 1 ]; then ! git status --porcelain | read || (git status; git diff; exit 1); fi
+  - if [[ "$PROTOBUF_VERSION" == "3.5.0" ]] && [[ "$TRAVIS_GO_VERSION" =~ ^1\.9\.[0-9]+$ ]]; then ! git status --porcelain | read || (git status; git diff; exit 1); fi
```
awalterschulze: 3.5.1 ?

awalterschulze:
That regex and the fact that I can't see git diff actually running is freaking me out :(

peter-edge:
What do you want to do?

awalterschulze:
Ah I see the go: 1.9.x has ruled out the possibility of simply doing an equal. Hmmm

awalterschulze:
I have checked previous versions that check diff and the print out is the same.  And now that I am back from holiday your regex looks really simple.

awalterschulze:
@peter-edge Thank you for putting up with these really drawn out comments.

PR #364, opened 2017-12-21
add bson tag for bson serialization, used for mongodb.

frankee:
When we need go struct, which protobuf generated, to support the bson serialization, within the gogoprotobuf, we can use `moretags`.

However in general, the value of `bson` tag is consistent with the `json` tag, so we only need to control whether the generator automatically adds `bson` tag by a switch, which defaults to false.

Currently, three switches have been added,
1. The generator CMD parameter
```
Protoc - gofast_out = bsontag:. My. Proto
```
2. The file option
```
Option (gogoproto bsontag_all) = true;
```
3. The message option
```
Option (gogoproto bsontag) = true;
```

The bson tag will check from the command line parameter to the message option, and the settings will override the previous one.

The specific value of the bson tag can be specified by `moretags`, eg. `moretags = "bson:\"fieldName\ "`, if `moretags` is not specified, then will automatically generate value for bson tag:

  1. First get the `json` attribute in `protobuf` tag, and if it exists, use the value;
  2. If the above does not exist, the value of the `json` tag is used.


awalterschulze:
Adding the command generator is a bit non conformant, and then it would suggest that we need to add a command generator for all tags.

This is what the vanity binaries are for.
If allows you to edit the tags before you send it to the generator.

For example https://github.com/gogo/protobuf/blob/master/protoc-gen-gogoslick/main.go

This will allow you to not need to create your own tag and simply use moretags to create bson tags for all fields, without the user having to specify them.

What do you think?

peter-edge:
See https://github.com/gogo/protobuf/pull/368

peter-edge:
@frankee can you take a look atg #368 and give your opinion? Especially regarding the `protobuf` vs `json` tag.

Issue #63, opened 2015-06-08
Vote For New Feature: Yaml `nullable=false` should generate `yaml:"omitempty"`


tamird:
This is currently done for JSON, but not YAML.

awalterschulze:
And what about bson and xml and all the other serialization formats.
I think this is what gogoproto.moretags is for.
We could add a vanity binary that supports yaml, but adding moretags by convention.
What do you think?

tamird:
I'd be 100% happy with a vanity function :)

awalterschulze:
Can we close this, or is anyone interested enough to do some work here?

tamird:
The JSON tag doesn't really matter - folks should be using jsonpb.

PR #367, opened 2018-01-11
Add YAML tags options

peter-edge:
This adds three options:

- A file option `yamltags_all` that will generate YAML tags that match the JSON tags
- A message option `yamltags` that does the same but just for the given message
- A field option `yamltag` that allows a custom YAML tag in the same manner as `jsontag`

I know this was discussed before in https://github.com/gogo/protobuf/issues/63, and it is totally valid to say "we only do JSON, other formats like BSON, XML, etc are on their own", but I think YAML is such a common case that it's worth natively supporting. Extra options are cheap (as shown by `gogoproto/gogo.proto`), it's opt-in, not opt-out, and hell I wouldn't be against supporting BSON and XML as special options either.

This would save me a *lot* of pain, I've dealt with not having YAML tags in generated Protobuf definitions for a while in not amazing ways by converting YAML to JSON and then unmarshalling into the Protobuf struct. I would love to be able to represent YAML configuration files as Protobuf messages, and these options would be a big help with that. Having to tag each field individually with `moretags` is too much work to use Protobuf messages for YAML configurations in general.

Note I didn't add more testing other than checking the generated file manually as I didn't want to add a test import on `gopkg.in/yaml.v2` without checking first. I actually don't mind just seeing the manual verification but that's just me.


awalterschulze:
My proposal here is to use moretags and yes creating them when they can be infered is painful. 
For this case I propose creating a vanity binary where you fill in moretags programmatically.
I really don't want to support every type of serialization format, especially in not in generator.go which is already hard to maintian, if this is already easy to do with a vanity binary.

peter-edge:
I get not wanting to support every type of serialization format, but YAML is almost on the same level as JSON, if not the exact same level. It really should be supported.

Creating vanity binaries is a fine idea in theory, but it's really hard to distribute that concept, especially when a lot of users are using `protoc-gen-gogo` within a company, both in internal code and in OSS. It's just not really feasible.

Tldr: YAML is basically equivalent to JSON, it's not a second-class citizen, and this would help in practice a ton for I bet more people than just me. I plead you to reconsider on this specific case :-)  BSON and XML are not in the same category as JSON and YAML for Golang developers.

awalterschulze:
JSON is actually supported by protobuf as a serialization format just like prototext and protobinary, since proto3, not in the tags which is another big problem, but this does make any other serialization format second class.

I struggle to understand why it is hard to distribute a different code generation binary inside a company:
```
protoc --gogo_with_yaml_out=.
```

peter-edge:
JSON is supported within the Protobuf tag, the extra JSON tag is not used by jsonpb for example.

It's just not great for consistency - sure, you could do it, and setup a whole company to use it (which for OSS would also mean creating a repo just for this, which is a mess), but an entire new program for what could be a simple option in gogo.proto?

I'm sure I am not the only person who would use this, this would be a big selling point for gogo in general. It's a great candidate for a feature in gogo natively, as opposed to requiring each person who wants YAML support to make their own binary and distribute it, and do it in their own way. It's going to lead to a lot of fragmentation (which there already is on YAML).

This is just such a common case that it should be included, in my opinion.

awalterschulze:
Another thing is that if I include this, then I have to test it.
And if I test it, then I have another dependency, which is a deal breaker for me.

The only reason I support the json tag is that you can't change it with more tags, since it was already included in golang/protobuf.

peter-edge:
I will write the tests for it

Additionally, test imports are not part of imports - if you only include 'gopkg.in/yaml.v2' in `_test.go` files, it will not be a dependency for gogo/protobuf on go get.

peter-edge:
I just thought of a compromise solution:

Things like YAML, BSON, XML effectively just want to use the JSON tag. What about two options:

- File option string copyjsontags_tall`
- Message option `string copyjsontags`

That you could set for example as:

```
option (gogoproto.copyjsontags_all = "yaml,bson";
```

That will copy what the json tag is into a new tag, ie if you have it set to yaml, it will mirror the json tag (which is what this is doing).

This is independent of the actual serialization format, and is the generic `moretags` case - you just want to copy the JSON tag into something else. So for people currently using `moretags` for xml, they could just set `copyjsontags_all` to `xml` and remove the `moretags` annotations.

How does this sound?

awalterschulze:
Now we are getting somewhere :)

Hmmm

Do we also capitalize the first letter for some and not others, I can't
remember?

peter-edge:
There's differing feelings about whether to use pascalCase or snake_case for both JSON and YAML, but there's no consensus. Protobuf just copies the field name I believe for JSON.

I think just doing what Protobuf does for JSON does makes this perfectly consistent. It even allows you to allow users to specify EITHER JSON or YAML config files.

Imagine:

```proto
syntax = "proto3";

package foo;

option (gogoproto.copyjsontags_all) = "yaml";

message Config {
  string log_level = 1;
}
```

And being able to either have a `config.json` or `config.yaml` that both use the same field declarations but different formats.

Two things:

1. I don't like the name `copyjsontags` that much, any other suggestions?
2. Should this be a `repeated string` option instead of a comma-separated `string` option?

awalterschulze:
I agree and I agree with 1. I will think about your questions.

Maybe let us also get some extra opinions from previous yaml feature
requesters. Just to make sure we are building the right thing.

peter-edge:
Lol I only just saw https://github.com/gogo/protobuf/pull/364, that's funny.

This would solve that case too.

peter-edge:
The only previous YAML thing I could find was #63, is there anything else?

awalterschulze:
I guess thats it.

awalterschulze:
We can also use the xml library for testing and not introduce another dependency :)

awalterschulze:
@frankee @tamird what do you think about Peter's proposal for copying json tags to other tags like yaml and bson.

peter-edge:
Closing this in favor of #368

PR #368, opened 2018-01-11
Add copyjsontagto, copyjsontagnameto, copyjsontagto_all, and copyjsontagnameto_all options

peter-edge:
Things like YAML, BSON, XML effectively just want to use the JSON tag. This adds four options:

- File options `string copyjsontagto_all, string copyjsontagnameto_all`
- Message option `string copyjsontagto, string copyjsontagnameto`

That you could set for example as:

```
option (gogoproto.copyjsontagto_all) = "yaml";
```

or:

```
option (gogoproto.copyjsontagto_all) = "yaml,bson";
```

That will copy what the json tag is into a new tag, ie if you have it set to yaml, it will mirror the json tag (which is what this is doing). If you use `*name` instead, it will not copy json options like `omitempty`, only using the field name.

This is independent of the actual serialization format, and is the generic `moretags` case - you just want to copy the JSON tag into something else. So for people currently using `moretags` for xml, they could just set `copyjsontagto_all` to `xml` and remove the `moretags` annotations.

https://github.com/gogo/protobuf/pull/367

One more note: there is a mismatch between the JSON field name in the `protobuf` tag, which uses pascalCase, and the JSON field name in thje `json` tag, which matches the field name on the message. So for example, if you have `optional foo_id = 1;` in your message, the `protobuf` tag will have this as `json=fooId`, and the `json` tag will have this as `json:"foo_id,omitempty"`. This is annoying and surely something we can't change, the question is what do we do for these copy tags. I think unfortunately we need to copy the `json` tag, although there's an argument to be made to match the `protobuf` tag as users should be using the jsonpb package for json, which is what this is for. **UPDATE:** I like what https://github.com/gogo/protobuf/pull/364 does, which is copy the `protobuf` tag if it exists, otherwise copy the `json` tag, this is an option.

There's also unfortunately an argument to be made that we shouldn't handle any of this, as serializing with something other than the internal serializers (binary, jsonpb) will result in something unexpected. jsonpb handles oneofs, well-known types, etc. This same problem already exists with using `moretags`, and users might just have to know this.

Questions to answer:

1. Better name than `copyjsontags`? `copyjsontagto`? So it would read `option (gogoproto.copyjsontagto) = "yaml";` **UPDATE:** using `copyjsontagto`
2. Should the option be comma-separated, or `repeated string`? **UPDATE:** This isn't allowed https://cxwangyi.wordpress.com/2012/04/13/mastering-google-protocol-buffer-custom-option-with-repeated-fields/
3. Do we want a second set of options, `copyjsontagname` and `copyjsontagname_all`, that omits JSON options ie omits `omitempty`? I think we want to copy `omitempty` in general, but in terms of having a generic tagging system, we might want to give people the option to just copy the json TAG name, and none of the options. **UPDATE:** added this


tamird:
I have no opinions on this, other than to say it is a shame that these commit messages do not document the rationale.

peter-edge:
The PR description does, when you squash a merge this is added. I think this is relatively standard.

tamird:
Eh, that has other problems like squashing unrelated changes like "copy gogo.pb.go to golden file" into a single commit. I don't want to get into a whole thing here, so do as you wish.

peter-edge:
I would think this whole thing should be one squashed commit, as the commits don't necessarily have meaning separately and include tests - this is generally how my team develops, and when we merge, we have the PR to link to to the single commit, and copy in the PR description.

awalterschulze:
Juts for reference:
Current logic for deciding the json

```go
jsonName := *field.Name
			jsonTag := jsonName + ",omitempty"
			repeatedNativeType := (!field.IsMessage() && !gogoproto.IsCustomType(field) && field.IsRepeated())
			if !gogoproto.IsNullable(field) && !repeatedNativeType {
				jsonTag = jsonName
			}
			gogoJsonTag := gogoproto.GetJsonTag(field)
			if gogoJsonTag != nil {
				jsonTag = *gogoJsonTag
			}
```

awalterschulze:
Questions, given the above logic:
1. Are we doing omitempty for other tags?
2. Are we also changing their tag names if the json tag name is changed?
3. Will people be confused with the json name that is in the protobuf tag.  What is the logic for that json name, since it comes from protoc.

peter-edge:
1. `omitempty` is added  using `copyjsontagto`, but not `copyjsontagnameto`, ie `copyjsontagto` uses the JSON logic, `copyjsontagnameto` just uses the name of the JSON tag and makes a new tag with no options.

2. If the JSON tag name is changed, yes, this is just copying the JSON tag, ie that's how it is named.

3. The difference beween the `json` name in the `protobuf` tag and the actual `json` tag is annoying, but that's just now it is as of now. See https://github.com/gogo/protobuf/pull/364 for another option that first looks for the  `protobuf` tag. I'm not sure which is preferable.

awalterschulze:
Another option would just be to have one extension called `nametag` which simply copies the field name.  But that assumes that omitempty is not a use case and it is feature of xml, bson and yaml, so I doubt that it won't be a use case.

But if we assume `nametag` then we could also have an `omitempty` extension, but then people would typically want both.

So copying the json tag probably still makes the most sense.

awalterschulze:
So is it even a use case not not include omitempty?  In other words could we get rid of copyjsontagnameto and only have copyjsontagto?

Don't do any coding yet.  I am still just discussing and thinking of how we can keep this as simple as possible.

peter-edge:
I think it is personally. We don't have to include it for now if that makes
more sense, but I can see just wanting to copy the tag with no options in
the generic case.

awalterschulze:
Ok but can you actually see the use case for not copying `omitempty`for a serialization format?  If these are in the minority, then these cases can still use moretags instead of copytag.  What do you think?

peter-edge:
I think it’s totally valid to not add a feature until it is needed, so not
adding copyjsontagnameto seems sensible.

peter-edge:
@awalterschulze I do wish we had the option to choose between the `protobuf:json=` tag and the `json:` tag. #364 wants the former, for example. Then I also go back and forth on the copy `omitempty` option. Ugh. Any thoughts? I think we're overthinking this.

awalterschulze:
There is a reason why I have been advocating for creating your own vanity
binary, so that I don't have to think about this.

Your idea was the first interesting one where I thought it might work, but
now we see it has its own complexities.

I want a clean solution. I don't think we are thinking about this enough,
yet.

peter-edge:
I'm not sure there's a "cleaner" solution - people want different things. We can support them all with a few options - which tag to use, whether to copy `omitempty`, etc.

awalterschulze:
If one of the cases is unpopular we can still refer to moretags and creating your own vanity binary.
I hope that one case is much more popular that the other.

peter-edge:
What should we do with this? Leaning towards closing.

awalterschulze:
Sorry I have been swamped lately.
I also don't know, this is a really tough issue.

peter-edge:
Closing this for now.

eaglerayp:
i believe this one is useful since many tag is similar to json tag.

alexi:
Is there any progress on this / any interest in re-opening? We are working on a system that stores lots of protobuf types in a mongo database and it would be great to have this implemented for `bson` tags. For what it's worth, `copyjsontagto` is the ideal solution for this use case.

matejkramny:
> Is there any progress on this / any interest in re-opening? I am working on a system that stores lots of protobuf types in a mongo database and it would be great to have this implemented for `bson` tags. For what it's worth, `copyjsontagto` is the ideal solution for this use case.

after some digging around, I found this quite useful: https://github.com/gogo/protobuf/pull/364#issuecomment-353297238

Example file for mongodb adding bson tags automatically. Posting to potentially help others using mongodb etc.

https://gist.github.com/matejkramny/35e519858f6e1eddee81780fc78cb208

PR #371, opened 2018-01-21
Added instructions for using proto files from google/protobuf

### LINKED ISSUE #370: add to readme that to use proto files from google/protobuf we need to add Mgoogle/protobuf/<filename>.proto=github.com/gogo/protobuf/types
Couldn't find this mentioned in obvious places and took a while of googleing before I found https://github.com/gogo/protobuf/blob/master/test/types/Makefile which shows the extra args I needed to set to get it to compile.

Also it would be nice if there's a github.com/gogo/protobuf/types/<filename>.proto we could import instead so we don't need to have a very long protoc command.

AlexRouSg:
Fixes #370


awalterschulze:
Nice Thank you :)

PR #377, opened 2018-02-08
Generate correct values in maps with enumvalue_customname

akaspin:
Fixed `XXX_name` and `XXX_value` maps generation for enum with `enumvalue_customname` extension.

```protobuf
enum MyEnumWithEnumStringer {
	STRINGER_A = 0 [(gogoproto.enumvalue_customname) = "EnumValueStringerA"];
	STRINGER_B = 1;
}
```
was:
```go
var MyEnumWithEnumStringer_name = map[int32]string{
	0: "STRINGER_A",
	1: "STRINGER_B",
}
var MyEnumWithEnumStringer_value = map[string]int32{
	"STRINGER_A": 0,
	"STRINGER_B":         1,
}
```
corrected:
```go
var MyEnumWithEnumStringer_name = map[int32]string{
	0: "EnumValueStringerA",
	1: "STRINGER_B",
}
var MyEnumWithEnumStringer_value = map[string]int32{
	"EnumValueStringerA": 0,
	"STRINGER_B":         1,
}
```

### DIFF
--- a/protoc-gen-gogo/generator/generator.go
+++ b/protoc-gen-gogo/generator/generator.go
@@ -1553,7 +1553,11 @@ func (g *Generator) generateEnum(enum *EnumDescriptor) {
 		if _, present := generated[*e.Number]; present {
 			duplicate = "// Duplicate value: "
 		}
-		g.P(duplicate, e.Number, ": ", strconv.Quote(*e.Name), ",")
+		name := *e.Name
+		if gogoproto.IsEnumValueCustomName(e) {
+			name = gogoproto.GetEnumValueCustomName(e)
+		}
+		g.P(duplicate, e.Number, ": ", strconv.Quote(name), ",")
 		generated[*e.Number] = true
 	}
 	g.Out()
@@ -1561,7 +1565,11 @@ func (g *Generator) generateEnum(enum *EnumDescriptor) {
 	g.P("var ", ccTypeName, "_value = map[string]int32{")
 	g.In()
 	for _, e := range enum.Value {
-		g.P(strconv.Quote(*e.Name), ": ", e.Number, ",")
+		name := *e.Name
+		if gogoproto.IsEnumValueCustomName(e) {
+			name = gogoproto.GetEnumValueCustomName(e)
+		}
+		g.P(strconv.Quote(name), ": ", e.Number, ",")
 	}
 	g.Out()
 	g.P("}")
--- a/test/combos/both/thetest.pb.go
+++ b/test/combos/both/thetest.pb.go
@@ -193,11 +193,11 @@ const (
 
 var YetAnotherTestEnum_name = map[int32]string{
 	0: "AA",
-	1: "BB",
+	1: "BetterYetBB",
 }
 var YetAnotherTestEnum_value = map[string]int32{
-	"AA": 0,
-	"BB": 1,
+	"AA":          0,
+	"BetterYetBB": 1,
 }
 
 func (x YetAnotherTestEnum) Enum() *YetAnotherTestEnum {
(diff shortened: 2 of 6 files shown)


awalterschulze:
This is not the use case.

From the documentation:
`enumvalue_customname : Changes the generated enum name. Helps with golint issues.`

It only changes the generated go enum name and is not supposed to change the string.
Changing this would require finding all users and asking them if changing the behaviour is ok.
I worry that this might also have an effect of the output jsonpb, which I assume would be very undesirable.

awalterschulze:
Why not change the enum name from `STRINGER_A` to `EnumValueStringerA`.  Then you get everything for free.

akaspin:
@awalterschulze 

> Why not change the enum name from STRINGER_A to EnumValueStringerA. Then you get everything for free.

Best example is https://github.com/google/openrtb/blob/master/openrtb-core/src/main/protobuf/openrtb.proto#L1899 - build fails with "proto3". It's pure pain.

> It only changes the generated go enum name and is not supposed to change the string.

Maybe introduce something like `enumvalue_mapname`?

awalterschulze:
Could you explain the problem more?

akaspin:
Like this:

```
openrtb/openrtb.proto:1032:9: When enum name is stripped and label is PascalCased (Iab11), this value label conflicts with IAB1_1. This will make the proto fail to compile for some languages, such as C#.
openrtb/openrtb.proto:1038:9: When enum name is stripped and label is PascalCased (Iab12), this value label conflicts with IAB1_2. This will make the proto fail to compile for some languages, such as C#.
...
```

I can rename values like `IAB1_c2`. But I also want to ensure compatibility with all dependent code.

awalterschulze:
I am really struggling to follow you sorry :(
I am doing many context switches.
So I am probably missing the obvious.

Can you create a small representative example with the error you are trying to solve.

akaspin:
Ok. Right now we have something like this:

```
syntax = "proto3";
...
enum ContentCategory {
...
	IAB1_1 = 2;      // Books & Literature
...
	IAB11 = 191;     // Law, Gov't & Politics
}
```
Latest `protoc` fails with:
```
openrtb/openrtb.proto:1032:9: When enum name is stripped and label is PascalCased (Iab11), this value label conflicts with IAB1_1. This will make the proto fail to compile for some languages, such as C#.
openrtb/openrtb.proto:1038:9: When enum name is stripped and label is PascalCased (Iab12), this value label conflicts with IAB1_2. This will make the proto fail to compile for some languages, such as C#.
...
```
For now all `.pb.go` files are generated with docker container with outdated protoc inside. I want to fix this.

But. I think this is wrong approach. Better solution is drill all dependent code and rename values.

awalterschulze:
Can you show the protoc command that fails?  which out and you using etc.

akaspin:
It's not plugin bug. It's `protoc` "feature" (https://github.com/google/protobuf/pull/2204). I'll make another PR for separate option.

awalterschulze:
Ah interesting. Did not know about that.

1. Why not stick to proto 2 and take your time to upgrade dependant code?

2. Or complain with Google protobuf to make it a warning again, since you
don't plan to use c#

3. Or you can try to turn off the generation of enum stringer and write
your own method and use that in combination with the enumvalue_customname.

akaspin:
@awalterschulze 

1. It's not possible because a lot of legacy dependant code uses generated stubs.
1. I really don't think that this is possible in reasonable time :-)
1. Choose this because on the side I found bug in production and need to use dashes instead underscores in maps.

awalterschulze:
Ok so number 3 works for you?

PR #381, opened 2018-02-14
protoc-min-version: don't suppress protoc out

skipor:
Have noticed, that protoc-min-version is suppressing warnings that protoc plugins write to stderr.
Seems to me, protoc-min-version should not change protoc behavior.
Also, it seems reasonable to `exit(1)` when the protoc version is not high enough.

### DIFF
--- a/protoc-min-version/minversion.go
+++ b/protoc-min-version/minversion.go
@@ -30,10 +30,11 @@ package main
 
 import (
 	"fmt"
-	"github.com/gogo/protobuf/version"
 	"os"
 	"os/exec"
 	"strings"
+
+	"github.com/gogo/protobuf/version"
 )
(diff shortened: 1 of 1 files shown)


awalterschulze:
I agree with warnings, but this is a tool used for testing on travis with
different protoc versions. Which means I can run all tests with any protoc
version. I need it to not exit 1 and this was the reason for creating this
binary.

If you have a different requirement it should be very simple to copy and
modify this binary's main.go

What do you think?

skipor:
Ok, I will use my modified version.
I have removed commit with `exit(1)` from PR.

awalterschulze:
Thanks, sorry for the inconvenience and thanks for this fix.

PR #393, opened 2018-03-15
Add to jsonpb.Marshaller option to flatten embedded messages

### LINKED ISSUE #211: Marshal embedded message into a flat JSON string
I'm trying to embed two messages into one while using the `jsontag` extension to allow the marshaler to marshal it into a flat JSON string.

I have the following messages in my `proto` file:

``` proto
message Meta {
    google.protobuf.Timestamp created_at = 1 [(gogoproto.stdtime) = true];
    google.protobuf.Timestamp updated_at = 2 [(gogoproto.stdtime) = true];
    google.protobuf.Timestamp deleted_at = 3;
}

message Person {
    string id = 1;
    Meta meta = 2 [(gogoproto.embed) = true, (gogoproto.jsontag) = ""];
}
```

In the generated `.pb.go` file, this is translated to types `Meta` and `Person` like this:

``` go
type Meta struct {
    CreatedAt *time.Time `protobuf:"bytes,1,opt,name=created_at,json=createdAt,stdtime" json:"created_at,omitempty"`
    UpdatedAt *time.Time `protobuf:"bytes,2,opt,name=updated_at,json=updatedAt,stdtime" json:"updated_at,omitempty"`
    DeletedAt time.Time  `protobuf:"bytes,3,opt,name=deleted_at,json=deletedAt,stdtime" json:"deleted_at,omitempty"`
}

type Person struct {
    Id    string `protobuf:"bytes,1,opt,name=id,proto3" json:"id,omitempty"`
    *Meta `protobuf:"bytes,2,opt,name=meta,embedded=meta" json:""`
}
```

But, the `jsonpb.Marshaler` produces the following JSON string:

``` json
{"id":"foo","meta":{"created_at":"2016-10-10T11:15:24.750978411Z"}}
```

While the [expected](https://play.golang.org/p/52wInuR8Jl) result is:

``` json
{"id":"foo", "created_at":"2016-10-10T11:15:24.750978411Z"}
```

It is worth mentioning that with the standard library JSON marshaler it works fine, but it limits us to use only primitive types. Other proto types like `StringValue`, `Int32Value`, etc. produce the same issue in the opposite way (because they are just a wrapper on top of the primitive type).

Attached a [gist](https://gist.github.com/liranp/6edeeb8179e7f1b263e34a48ef2a3523) with the relevant files.

Any help will be much appreciated!

dotpy3:
This PR closes #211, by adding an option to flatten embedded messages when (un)marshalling with `jsonpb`.

This option can be enabled with a new boolean flag in the marshaller, `jsonpb.Marshaller.EmbedAsStdJSON`. When this flag is enabled, if a struct field is `embed` and follows the `encoding/json` conditions to be flattened (no JSON tag or name), it is flattened.

For example, here are Protobuf messages that were added to the test suite:

```proto
message Town {
  required string name = 1;
  required string postcode = 2;
}

message Address {
  required Town town = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false, (gogoproto.jsontag) = ""];
  required string street = 2;
}
```

When marshalling an `Address` with `jsonpb.Marshaller` with this `EmbedAsStdJSON` flag enabled, the marshaller will return messages in this format:

```json
{"postcode": <postcode>, "street": <street>, "name": <name>}
```

Whereas it used to marshall it to the following format:

```json
{"town":{"postcode": <postcode>, "name": <name>}, "street": <street>}
```

No flag was added to the unmarshaller: the unmarshaller can also now unmarshal both flattened and non-flattened JSONs.


dotpy3:
Realized that my first commit did not handle optional values, so I added that in a second commit, along with tests to verify it. I'm not sure why CI is unsuccessful, did you already face that issue with GopherJS?

awalterschulze:
I have fixed the GopherJS problems on master.

awalterschulze:
I think this is a hard to maintain diff.
Wouldn't it be easier to write/generate a json marshal method for this struct, that enables the modified behaviour?

I am also worried about how this will play with grpc-gateway or how API users from other languages, given the .proto file will be able to consume this API?  Since the users from other languages, won't be able to unmarshal this json.

dotpy3:
> I think this is a hard to maintain diff.
> Wouldn't it be easier to write/generate a json marshal method for this struct, that enables the modified behaviour?

It's a consequent diff indeed - let me try and look if I can make it easier to maintain!

> I am also worried about how this will play with grpc-gateway or how API users from other languages, given the .proto file will be able to consume this API? Since the users from other languages, won't be able to unmarshal this json.

Indeed yes, it might not play well with other languages that don't have a similar feature if they're using the native proto<->JSON marshaller. Our specific use case is to make our JSON API (built using `grpc-gateway`) nicer to use - but we're keeping our gRPC endpoint open for users of other languages.
For other `gogo/protobuf` users, it would still be an opt-in option. Maybe we could document the feature better, so that those interested in the feature would be warned about the risks?

awalterschulze:
How hard would it be to write a marshal and unmarshal method for that
struct that gives you the same behavior?

dotpy3:
In our specific case, we'd like to use `embed` for our device IDs across our application:

```proto
message DeviceID {
  // LoRaWAN JoinEUI
  bytes join_eui = 1;
  // LoRaWAN DevAddr
  bytes dev_addr = 2;
}

message Entry1 {
  // Device
  DeviceID device = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false];
  // ...data, such as:
  string name = 2; // Name of the device
}

message Entry2 {
  // Device
  DeviceID device = 1 [(gogoproto.embed) = true, (gogoproto.nullable) = false];
  // ...data
}
```

An device can either be identified by its `JoinEUI` or by its `DevAddr`, hence the choice of using an intermediate structure rather than using `bytes` directly as key in `Entry1` and `Entry2`. On the other side, we'd like to enable API users to make clean, flattened JSON calls like this:

```json
{
    "devAddr": "2A018242",
    "name": "Temperature sensor"
}
```

... rather than making users add a `DeviceID` object in the JSON message.

In this situation, we'd have to write custom marshallers not for `DeviceID`, but for every `Entry` object using `DeviceID` as an identifier. I think that it would create us a lot of technical debt, since every new object in our application referencing a Device would need a custom marshaller. Having the possibility to flatten JSON as an option of the marshaller would help us (and other `protobuf` users) save a lot of code and time spent maintaining those marshallers!

awalterschulze:
Ok that makes sense.  A custom json marshaler and unmarshaler is not going to solve that unfortunately.

How about a function that flattens and unflattens your marshaled json structures
https://gist.github.com/awalterschulze/86401e1080dfa2d7fc86ed0991146b52
I assume that when users are using json over protobufs, that they are not that interested in speed. But is so then the parser is semi optimized and I am keen on making it faster and there is no reason the encoding functions cannot be optimized either.

I am really trying to avoid to many modifications to jsonpb.  This package has a lot of issues and will continue to have.  Maintaining it will be even worse if I start changing it.  So I am really hoping you can find another solution for this case.

dotpy3:
Thank you for taking the time to make such a constructive response!
I understand your concerns about maintenance of jsonbp. I think that on our side, after spending more thought on this, we've accepted to have `"id": {...}` objects within the JSON, so the final solution we found doesn't require flattening anymore. I'm closing this pull request - but thank you for looking into it!

awalterschulze:
Glad you found something that works for you.  Really sorry we had to take such a round about way.

PR #399, opened 2018-04-05
Upstream

awalterschulze:
Merging upstream from golang/protobuf into dev branch

https://github.com/golang/protobuf/commit/8cc9e46429bfb16289d40d30b2ee3f4923b47345


sitano:
merging this into master broke backward compatibility as described in https://github.com/golang/protobuf/issues/607 . before that those XXX_ fields were disablable in protoc-gogo-. now it's seems like not.

PR #410, opened 2018-05-30
feat(enum): add omit type alias option for generated constants

acrazing:
Usage:

```protobuf
syntax = "proto3";

enum Gender {
  option (gogoproto.enum_omittypealias) = true;
  Unknown = 0;
  Female = 1;
  Male = 2;
}
```

This will emit code like this:

```go
type Gender int32

const (
  Code_Unknown = 0
  Code_Female = 1
  Code_Male = 2
)
```


awalterschulze:
What is the use case ?

acrazing:
For example: we use protoc generate global `gRPC` status code as follow:

```protobuf
enum Code {
  Ok = 0;
  UsedAlready = 1;
}
```

This option allow us use the generated code as the parameter of the function `status.Error` directly, rather than cast it by `codes.Code(x.UserAlready)`.

The reason for Using protoc to generate status codes is that it could check all the code is unique.

awalterschulze:
Couldn't you also use go's iota to achieve the same thing?

PR #429, opened 2018-07-17
consistent import naming between the import declaration and the vars.…

### LINKED ISSUE #427: Since `dev` merge, code no longer compiles, lots of strange changes
Over the weekend, it would appear that much of the generated code no longer compiles, almost all of which is related to the packages `golang.org/x/net/context` (I'm using Go 1.10.3) or `google.golang.org/grpc` not being used:

```
./cluster.pb.go:18:8: imported and not used: "golang.org/x/net/context" as golang_org_x_net_context
./cluster.pb.go:19:8: imported and not used: "google.golang.org/grpc" as google_golang_org_grpc
```

If I run `goimports -w` on the generated files it does fix these errors, but there appear to be a large number of strange changes. for instance:

* All the `fileDescriptor` names moved from (e.g.) `fileDescriptorCluster` to `fileDescriptor_cluster_c98421a8c26a96a1` (making diffs a lot less readable)
* There are a number of strange package import aliases, such as `import strings "strings"` and `import reflect "reflect"` etc -- do these really need to be aliased?
* Some of the import aliases have been renamed to strange unwieldly strings, e.g. `import sortkeys "github.com/gogo/protobuf/sortkeys"` is now `import github_com_gogo_protobuf_sortkeys "github.com/gogo/protobuf/sortkeys"` -- is that deliberate?
* Using `gogofaster` or `gogoslick` there are now a lot of `XXX_` prefixed exported variables in structs, such as `XXX_sizecache` and `XXX_NoUnkeyedLiteral`, and the re-appearance of methods such as `XXX_Unmarshal`, `XXX_Marshal`, `XXX_Merge`, `XXX_Size`, `XXX_DiscardUnknown` etc.

I can't pinpoint where these issues began within the dev merge, but something is definitely substantially different compared to Friday. Should I be following some particular branch for a stable version?

jmarais:
… This is more consistant with golang/protobuf. fix #427


jmarais:
Ah.
"cannot find package "golang.org/x/net/context""
"cannot find package "google.golang.org/grpc""
Do we want some other tests that could check for consistent import referencing?

awalterschulze:
Sorry for the delay

What we want is to not include grpc or any other dependencies in the repo, but still test this.
I have had a similar problem with testing gofast_out which requires golang/protobuf here

https://github.com/gogo/protobuf/blob/master/Makefile#L143-L144
https://github.com/gogo/protobuf/blob/master/Makefile#L146
https://github.com/gogo/protobuf/blob/master/vanity/test/Makefile#L43-L46
https://github.com/gogo/protobuf/blob/master/vanity/test/gofast/.gitignore

Could we do something similar for this problem?

test/issue427/Makefile
```diff
@@ -1,4 +1,7 @@
-regenerate:
+test:
```
awalterschulze: Can we move this to only be under the `testall` target.
I don't want the `test` target to download anything.

And could we move the protoc-min-version command into a Makefile in test/issue247?

awalterschulze: my bad, wrong diff

Issue #436, opened 2018-07-24
Unmarshal allocates twice the memory needed for repeated fields


gunnihinn:
The generated code that deserializes a repeated field is essentially:

    xs := make([]type, 0)
    for _, x := range things {
        xs = append(xs, x)
    }

When the Go runtime needs to grow the backing array of a slice, it uses a doubling strategy to do so.  If we have _n_ elements to append to an empty slice, and _k_ is such that _2^k < n < 2^{k+1}_, then this means we usually end up allocating space for at least

    1 + 2 + 4 + ... + 2^{k+1} = 2 * 2^{k+1}

elements, while making a single allocation of _n_ elements would have been enough.

When we deserialize a protobuf message with a packed repeated field, we know how many elements we're going to deserialize and can allocate all the memory needed in one go. [Here](https://github.com/gunnihinn/protobuf-alloc) is a demonstration of a simple program that does this. Profiling its memory use with the default generated code, and also with a single ad-hoc patch that performs this allocation, shows that what we describe above really does happen.

I work for Booking.com, and we're very interested in making this deserialization code more clever about its allocations if possible and acceptable to the maintainers. We use this library in our implementation of a [Graphite server](https://github.com/go-graphite/carbonapi), and Unmarshal is responsible for the vast majority of its memory use (around 20 GB when things are calm), which is unsurprising as the server spends most of its time deserializing very large arrays of doubles.  Getting rid of this behaviour would cut our memory use by about half.

## Sample results from running the [above program](https://github.com/gunnihinn/protobuf-alloc):

```
## Commit 36a0ed7 (current behaviour):

$ ./protobuf-alloc
$ go tool pprof protobuf-alloc memory.prof
protobuf-alloc
Type: inuse_space
Time: Jul 24, 2018 at 12:01pm (CEST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top                                                                                                                                                                  
Showing nodes accounting for 5175.25MB, 100% of 5175.25MB total                                                                                                              
      flat  flat%   sum%        cum   cum%
 2871.24MB 55.48% 55.48%  2871.24MB 55.48%  github.com/gunnihinn/protobuf-alloc/foo.(*Foo).Unmarshal
    2048MB 39.57% 95.05%  5175.25MB   100%  main.main
  256.01MB  4.95%   100%   256.01MB  4.95%  github.com/gunnihinn/protobuf-alloc/foo.(*Foo).Marshal
[...]

PR #437, opened 2018-07-25
Gmagnusson/exact slice allocation

gunnihinn:
See comments in commit message and [issue 436](https://github.com/gogo/protobuf/issues/436).


gunnihinn:
I thought a bit more about the throwaway comment about packed types of non-fixed length. I think it's worth a shot to do that, and tune it in such a way that maybe it only kicks in if we're sure there are enough varints to deserialize to be worth it. It sounds like a nice project for junior developers, so I'm fishing for some at $work to see if anyone wants to have a go at this. It might take a little while to organize, so I think we'll do a separate PR for that.

awalterschulze:
Thank you so much @gunnihinn this looks very promising.

@donaldgraham would you mind reviewing this?

donaldgraham:
This looks great.  I like the way the test expresses the expectation that we allocate less than half of the memory we had allocated before...

Issue #438, opened 2018-07-25
Undefined ProtoSize() method when using google.protobuf.Any 


jwilander:
As part of a larger project, I have some messages defined like so:

```proto
message API_LoadPluginConfigurationRequest {
	google.protobuf.Any arg1 = 1;
}
```

The `MarshalTo` method that gets generated is:

```Go
func (m *API_LoadPluginConfigurationRequest) MarshalTo(dAtA []byte) (int, error) {
	var i int
	_ = i
	var l int
	_ = l
	if m.Arg1 != nil {
		dAtA[i] = 0xa
		i++
		i = encodeVarintPlugin(dAtA, i, uint64(m.Arg1.ProtoSize()))
		n45, err := m.Arg1.MarshalTo(dAtA[i:])
		if err != nil {
			return 0, err
		}
		i += n45
	}
	if m.XXX_unrecognized != nil {
		i += copy(dAtA[i:], m.XXX_unrecognized)
	}
	return i, nil
}
```

Giving me the error:

```
m.Arg1.ProtoSize undefined (type *types.Any has no field or method ProtoSize)
```

I'm including the following options when running `protoc`:

```
--gofast_out=plugins=grpc,Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/duration.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/timestamp.proto=github.com/gogo/protobuf/types:$(GOPATH)/src
```

This happens for other messages as well. Some with repeated Any types or Any types used as values in maps.

I'm not sure if this is a bug or something I'm doing wrong. If it is a bug I'd be glad to help to track down and fix it, though I'm still new to protocol buffers. Any ideas?

awalterschulze:
Could you also include a small example of a proto file that uses any, that
reproduces this problem?

Sounds like this might be a bug on our side.

jwilander:
Sure, this subset of my .proto file will reproduce it. I imagine it has something to do with the options being specified, though I (possibly incorrectly) assume it should work regardless of how those are set

```proto
syntax = "proto3";
package mmproto;

import "github.com/gogo/protobuf/gogoproto/gogo.proto";
import "google/protobuf/any.proto";

option (gogoproto.protosizer_all) = true;
option (gogoproto.sizer_all) = false;
option go_package = "mmproto";

message API_LoadPluginConfigurationRequest {
	google.protobuf.Any arg1 = 1;
}

[...]

PR #443, opened 2018-08-01
 Added ProtoSize wrapper functions for the well known types  #438

jmarais:
@awalterschulze 
Is there a way to generate the Value_** types? I tried, but failed.


awalterschulze:
@jmarais 
Interesting, they seem to have generated in wrappers.pb.go
https://github.com/gogo/protobuf/blob/master/types/wrappers.pb.go

Here they are used
https://github.com/gogo/protobuf/blob/ebc05652de1e3bd0da5f0879f2548d6580eee8fa/jsonpb/jsonpb_test_proto/test_objects.proto#L150

So this makefile should show you have to generate them
https://github.com/gogo/protobuf/blob/ebc05652de1e3bd0da5f0879f2548d6580eee8fa/jsonpb/jsonpb_test_proto/Makefile

What is the error message you are getting?

jmarais:
It seems like most of the "Values" are fine. Only NumberValue and StructValue gives a problem:

```bash
issue438.proto:34:3: "google.protobuf.NumberValue" is not defined.
issue438.proto:37:3: "google.protobuf.StructValue" is not defined.
```
I see that they are not defined in wrappers.pb.go

jmarais:
Ok. So I can change the StructValue to Struct. And leave out the NumberValue.

jmarais:
Hmm. I think we are generating the NullValue wrong.

```golang
func (m *Types) GetNlval() types.NullValue {
	if m != nil {
		return m.Nlval
	}
	return types.NullValue_NULL_VALUE
}
```

```bash
./issue438.pb.go:157:9: undefined: types.NullValue_NULL_VALUE
```

There is only this const defined in struct.pb.go:

```golang
const (
	// Null value.
	NULL_VALUE NullValue = 0
)
```

jmarais:
It is this line causing the wrong enum name:
https://github.com/gogo/protobuf/blob/master/protoc-gen-gogo/generator/generator.go#L2686

awalterschulze:
Good catch

So maybe we should remove this line?
https://github.com/gogo/protobuf/blob/master/protoc-gen-gogotypes/main.go#L47

jmarais:
That seems to work.
It generates:
```golang
const (
	// Null value.
	NullValue_NULL_VALUE NullValue = 0
)
```
in struct.pb.go.

Do we want to make that change? I am unsure what else it will affect (if anything).

awalterschulze:
I think lets make the change and see the diff :)

What do you think?

types/type.pb.go
```diff
@@ -32,9 +32,9 @@ type Syntax int32
 
 const (
 	// Syntax `proto2`.
-	SYNTAX_PROTO2 Syntax = 0
+	Syntax_SYNTAX_PROTO2 Syntax = 0
```
awalterschulze: Here is the one other change

types/type.pb.go
```diff
@@ -47,51 +47,51 @@ var Syntax_value = map[string]int32{
...
+	return fileDescriptor_type_0082d870c49329d7, []int{0}
 }
 
 // Basic field types.
 type Field_Kind int32
 
 const (
 	// Field type unknown.
-	TYPE_UNKNOWN Field_Kind = 0
+	Field_TYPE_UNKNOWN Field_Kind = 0
```
awalterschulze: Oh and here

types/type.pb.go
```diff
@@ -138,21 +138,21 @@ var Field_Kind_value = map[string]int32{
...
+	return fileDescriptor_type_0082d870c49329d7, []int{1, 0}
 }
 
 // Whether a field is optional, required, or repeated.
 type Field_Cardinality int32
 
 const (
 	// For fields with unknown cardinality.
-	CARDINALITY_UNKNOWN Field_Cardinality = 0
+	Field_CARDINALITY_UNKNOWN Field_Cardinality = 0
```
awalterschulze: more

awalterschulze:
I think given those are the only changes, then it is ok to generate the enum prefixes.

awalterschulze:
@johanbrandhorst are these enum prefixes for well known types affecting anything in grpc?

jmarais:
Yeh. Currently those values are only used in the /types package.

johanbrandhorst:
:grimacing: I'd hate to think anything hardcodes the enum values, I think all enums are looked up from the proto registry.

awalterschulze:
Ok then I think we are ready to merge :)
@jmarais are you ready?

jmarais:
Yup. Sounds good.

PR #445, opened 2018-08-03
Fix wrong build tags

s-hashimoto00:
Fix wrong build tags.

Currently:
```
$ go test ./proto/
ok      github.com/gogo/protobuf/proto  0.025s
$ go test -tags=purego ./proto/
proto/pointer_unsafe_gogo.go:40:6: pointer.getRef redeclared in this block
        previous declaration at proto/pointer_reflect_gogo.go:43:6
proto/pointer_unsafe_gogo.go:41:18: unknown field 'p' in struct literal of type pointer
proto/pointer_unsafe_gogo.go:44:6: pointer.appendRef redeclared in this block
        previous declaration at proto/pointer_reflect_gogo.go:47:6
proto/pointer_unsafe_gogo.go:51:6: pointer.getSlice redeclared in this block
        previous declaration at proto/pointer_reflect_gogo.go:54:6
FAIL    github.com/gogo/protobuf/proto [build failed]
```

After fixes:
```
$ go test ./proto/
ok      github.com/gogo/protobuf/proto  0.024s
$ go test -tags=purego ./proto/
ok      github.com/gogo/protobuf/proto  0.048s
```

### DIFF
--- a/proto/pointer_unsafe_gogo.go
+++ b/proto/pointer_unsafe_gogo.go
@@ -26,7 +26,7 @@
 // (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-// +build !purego !appengine,!js
+// +build !purego,!appengine,!js
 
 // This file contains the implementation of the proto field accesses using package unsafe.
 


awalterschulze:
Great catch :)

I restarted the job, it looks like it should succeed.

Is it possible contribute a test, maybe adding something to the Makefile that tests that we don't make this mistake again?

awalterschulze:
I think we also don't support appengine/purego for non generated code anyway.
So if this could be the start of testing this and getting more support towards finally fixing this, it would be great.

s-hashimoto00:
Thanks for response.
I trying add test for appengine and purego.

awalterschulze:
Thanks for reporting and fixing and now contributing.
It is really appreciated :)

s-hashimoto00:
Updated:

Add tests for purego and appengine.
Some tests are skipped because of known issues.

- #447 test failed for jsonpb (upstream issue)
- #448 unable to casttype extension
- #449 unable to nullable extension

awalterschulze:
Good job, this looks great :)

It does increase the total build time from 26min tot 40min
Maybe it is good enough to only run the tests for `purego`
And maybe we can make a separate makefile target similar to `js` that is run as part of the `buildserverall` target

What do you think?

s-hashimoto00:
Thank you for review.
Fixed the makefile (force push).

- Only run tests for `purego` build tags
- Append `purego` target, and append `purego` to `buildserverall` target

PR #446, opened 2018-08-06
merged commit 32a84b27e28ab9f681f0df16160c4ef1f6587094 from golang/pr…

donaldgraham:
…otobuf

Please check if you're happy with the commit message.


awalterschulze:
It looks good except for the commit message being cut off `…otobuf`

awalterschulze:
The fail you are getting is only because you need to regenerate all files.

Running `make all` should fix this.

PR #457, opened 2018-08-10
WIP: Memory pooling

pedgeio:
This is the initial work for memory pooling, wanted to put it up before I spent too much time on it so we could look it over. Tldr for a message with a few fields and a repeated field of size 10, 1664B and 33 allocs/op -> 0B and 0 allocs/op.

Check out the `test/pool` directory.

Note there is no change for existing users. You have to enable the `(gogoproto.pool)` option, and even if you enable it, there is no functional change in in terms of current usage - messages can still be allocated with or without a backing Pool.

Some caveats:

- String and byte slice fields are not pooled. Strings basically never could be since they are immutable, byte slices in theory could, but it would cause a world of pain and not be backwards compatible to try to pool them when they are fields. I added a `MarshalPool` method that does do byte slice pooling using a seg list with a wrapper `*mem.Bytes` type, however.

A lot has changed since this PR was first put up. See the discussion below for the current state of this PR.


johanbrandhorst:
This is a super cool idea, I haven't read through the design yet, but perhaps to make it easier to see the speedup, can you publish benchmarks compared to _not_ using the pooling?

mem/mem.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+// Truncate truncates the value to the given length.
+func (b *Bytes) Truncate(valueLen int) {
+	b.valueLen = valueLen
+}
+
+// MemsetZero sets the data inside the value to zero.
+//
+// Optimized per https://golang.org/cl/137880043
```
johanbrandhorst: TIL :heart_eyes:

mem/mem.go
```diff
@@ -0,0 +1,325 @@
...
+//
+// By default, use DefaultChannelSize. Set this to 0 to not use any channel.
+func WithChannelSize(channelSize uint16) PoolOption {
+	return func(pool *Pool) {
+		pool.channelSize = channelSize
+	}
+}
+
+// WithSegListSizes returns a PoolOption that sets the seg list sizes
+// for allocating Bytes.
```
johanbrandhorst: I don't know what `seg list sizes` means. Could you elaborate a little in this comment? What compromises would the user make by changing this?

mem/mem.go
```diff
@@ -0,0 +1,325 @@
...
+// PoolOption is an option when constructing a new Pool.
+type PoolOption func(*Pool)
+
+// WithChannelSize returns a PoolOption that sets the Pool channel size.
+//
+// If your workflow usually only has one object of a given type allocated,
+// the garbage collector will generally garbage collect the object if it
+// is in a sync.Pool before it can be allocated again from the Pool. If this
+// option is used, a channel of allocated PooledMessages of this buffered
+// size will be used to recycle PooledMessages before using the sync.Pool.
```
johanbrandhorst: Even though this is neat, I, as a user, still am not clear on what consequences this has for my program. Could you give a practical example in this comment?

mem/mem.go
```diff
@@ -0,0 +1,325 @@
...
+}
+
+// WithNoLocking returns a PoolOption that turns off locking around message constructor
+// registration. If this option is turned on, you must make all RegisterConstructor calls
+// at initialization, and not call RegisterConstructor afterwards.
+//
+// Generally, you should use this option as it provides a significant performance benefit,
+// just make sure to only call registration functions at initialization.
+//
+// This option is called for the Global Pool.
```
johanbrandhorst: I think `The Global Pool has this option enabled already.` is a little more explicit.

mem/mem.go
```diff
@@ -0,0 +1,325 @@
...
+//
+// This generally is meant for proto.Messages but is compatible with
+// anything that has a Reset and Recycle method.
+type PooledMessage interface {
+	Reset()
+	Recycle()
+}
+
+// Pool is a pool of PooledMessages and Bytes.
+type Pool struct {
```
johanbrandhorst: Document explicitly if methods on the pool are generally thread safe (I get the impression they are?).

mem/mem.go
```diff
@@ -0,0 +1,325 @@
...
+
+// Truncate truncates the value to the given length.
+func (b *Bytes) Truncate(valueLen int) {
+	b.valueLen = valueLen
+}
+
+// MemsetZero sets the data inside the value to zero.
+//
+// Optimized per https://golang.org/cl/137880043
+func (b *Bytes) MemsetZero() {
```
johanbrandhorst: In the above linked CL this is referred to as `MemClr`. Do we want to follow that precedent?

mem/message_pool.go
```diff
@@ -0,0 +1,81 @@
+// Protocol Buffers for Go with Gadgets
+//
+// Copyright (c) 2013, The GoGo Authors. All rights reserved.
```
johanbrandhorst: 2018

mem/seg_list_pool.go
```diff
@@ -0,0 +1,91 @@
+// Protocol Buffers for Go with Gadgets
+//
+// Copyright (c) 2013, The GoGo Authors. All rights reserved.
```
johanbrandhorst: 2018

plugin/pool/pool.go
```diff
@@ -0,0 +1,246 @@
+// Protocol Buffers for Go with Gadgets
+//
+// Copyright (c) 2013, The GoGo Authors. All rights reserved.
```
johanbrandhorst: 2018

awalterschulze:
But please keep us up to date with your other experimental benchmarks.  I think everyone is very interested to see if this can be another major improvement.

@jmarais @donaldgraham

protoc-gen-gogo/generator/generator.go
```diff
@@ -3366,6 +3384,64 @@ func (g *Generator) generateExtensionRegistration(ext *ExtensionDescriptor) {
...
+	if field.IsRequired() || gogoproto.NeedsNilCheck(message.proto3(), field) {
+		g.P(`m.`, fieldName, ` = nil`)
+		return
+	}
+
+	var zeroValue string
+	switch *field.Type {
+	case descriptor.FieldDescriptorProto_TYPE_DOUBLE,
+		descriptor.FieldDescriptorProto_TYPE_FLOAT:
+		zeroValue = "0.0"
```
johanbrandhorst: Is this distinction from other numbers necessary? I figured this would be an untyped literal.

test/pool/Makefile
```diff
@@ -0,0 +1,41 @@
+# Protocol Buffers for Go with Gadgets
+#
+# Copyright (c) 2013, The GoGo Authors. All rights reserved.
```
johanbrandhorst: 2018

test/pool/pool_test.go
```diff
@@ -0,0 +1,205 @@
+// Protocol Buffers for Go with Gadgets
+//
+// Copyright (c) 2013, The GoGo Authors. All rights reserved.
```
johanbrandhorst: 2018

test/pool/poolfalse/poolfalse.proto
```diff
@@ -0,0 +1,50 @@
+// Protocol Buffers for Go with Gadgets
+//
+// Copyright (c) 2013, The GoGo Authors. All rights reserved.
```
johanbrandhorst: 2018

test/pool/pooltrue/pooltrue.proto
```diff
@@ -0,0 +1,51 @@
+// Protocol Buffers for Go with Gadgets
+//
+// Copyright (c) 2013, The GoGo Authors. All rights reserved.
```
johanbrandhorst: 2018

peter-edge:
The big issue I'm having right now is the use of lookups with the string message type, ie `Get("pooltrue.Foo")`, `Put("pooltrue.Foo", foo)`, etc. `runtime.mapaccess2_faststr` is accounting for **12 (TWELVE) percent** of the total time. I made this 10% faster by adding a pre-cached `GetMessagePool` option in the `init` function that uses the global.

I just put this in the description above:

_Update: I think we should do only globals. We should all get on a chat. There's some optimizations that while technically possible without a global pool, make the API so much less intuitive that it won't be adopted. I think this is maybe the only case where a (replaceable) global makes sense - the garbage collector is a "global" in a sense._

If I have only globals, AND I add functionality to go from `FieldDescriptorProto -> FileDescriptorProto` - which shouldn't be that hard, just need to make a wrapper `generator.FieldDescriptor` similar to `generator.FileDescriptor, generator.Descriptor` and then put the `*generator.Descriptor` on it - then there's some optimizations I can do that require **no maps, no casting**, but also while _not making the API unusable_. I'm really sensitive to making sure the API is intuitive, it's one of the reasons I'm not crazy about FlatBuffers in Golang, for example.

The entire implementation would be rocket fast (to the point where I expect no overhead for this code at all, knock on wood), and the only change users would make is using `pkg.GetFoo()` instead of `&pkg.Foo{}` to create messages. Now that I have global disable and enable of the global pool, this makes it so you can switch all your calls to `GetFoo` with **no effect** on your current code, and then when you want to enable pooling, call `mem.Global().Enable()`, and the only change you have to make is to call `Recycle()` properly.

By doing this, I can also update the gRPC plugin to *automatically do this server-side* so that all incoming messages are always pooled, with *no change to existing implementations*. Huge win for all existing codebases.

Any preliminary thoughts? If everyone is on board, I want to go the global-only route. We can all also hop on a chat/video chat to talk about this, I'm available this weekend.

jmarais:
I may be wrong, but this does not support pooling when your message contains a well known type. I added a any field to test/pool/pooltrue/pooltrue.proto
```protobuf
message Foo {
  uint64 one = 1;
  fixed32 two = 2;
  Bar bar = 3;
  repeated Bar repeated_bar = 4;
  map<uint32, Bar> map_bar = 5;
  google.protobuf.Any any = 6;
}
```
regenerated with:
```bash
protoc --proto_path=../../../../../:../../protobuf/:. --gogo_out=Mgoogle/protobuf/any.proto=github.com/gogo/protobuf/types:. pooltrue/pooltrue.proto
```
and just ran the test:
```bash
go test -v -run=. -bench=.
```
and got these errors:
```
# github.com/gogo/protobuf/test/pool/pooltrue
pooltrue/pooltrue.pb.go:450:6: iface redeclared in this block
        previous declaration at pooltrue/pooltrue.pb.go:434:6
pooltrue/pooltrue.pb.go:844:27: impossible type assertion:
        *types.Any does not implement mem.PooledMessage (missing Recycle method)
```

It seems like the Recycle method is generated like this:
``` golang
func (this *Foo) Recycle() {
	if this == nil {
		return
	}
	var iface interface{} = this.Bar
	if pooledMessage, ok := iface.(github_com_gogo_protobuf_mem.PooledMessage); ok {
		pooledMessage.Recycle()
	}
	for _, elem := range this.RepeatedBar {
		var iface interface{} = elem
		if pooledMessage, ok := iface.(github_com_gogo_protobuf_mem.PooledMessage); ok {
			pooledMessage.Recycle()
		}
	}
	for _, elem := range this.MapBar {
		var iface interface{} = elem
		if pooledMessage, ok := iface.(github_com_gogo_protobuf_mem.PooledMessage); ok {
			pooledMessage.Recycle()
		}
	}
	var iface interface{} = this.Any
	if pooledMessage, ok := iface.(github_com_gogo_protobuf_mem.PooledMessage); ok {
		pooledMessage.Recycle()
	}
	if this.pool == nil {
		return
	}
	this.pool.Put("pooltrue.Foo", this)
}
```
And then the well known types doesn't implement the ```Recycle``` method.

I didn't see this in your "Remaining work:" section.

peter-edge:
@awalterschulze just had a video call, putting some thoughts down:

- We both hate globals. This is always going to be true.
- Because Golang doesn't have the concept of Thread ID, and because goroutines can number in the tens of thousands, a mutex somewhere for the pool is inevitable, whether it be through a mutex, a channel, or something of the sort. There's no concept of Thread-Local Storage we can do.
- If you don't use globals here, there's two options.

One, you can have a universal Pool object that can store different message types in a map. This map shouldn't be string-based, you probably want either a pointer type assertion, or assign integer ids for every message type. Regardless, map lookups have a non-trivial overhead, and then you run into a "global lock problem", where you either need to lock adding new message types to this map, which means you need a read lock around accessing this map, which in my testing has a significant (10%) performance hit, or you need users to make sure they register every message type ahead of time in this (now non-global-variable) Pool, which can lead to issues at an organizational level when users forget to call this registration.
- Two, you can have Pool types per message, and pass these Pool types around. This will lead to a very heavy API.

- Memory management will be hard enough to get right with users calling Recycle properly.

We also discussed the need to have a variable on each Message to track memory management and have the Recycle method, versus no extra variables and having `pkg.PutFoo`. Regardless of `Recycle` vs `PutFoo`, you're going to want a variable on each Message to track whether the message has been recycled but not reset, and panic if the message is used in this case, because otherwise you risk users unknowingly using recycled messages, which is one of the worst types of bugs.

As I put it, you have three properties: global, messy, fast. You can do global and fast, non-global and messy, or non-global and slow. Global might be the best option - you don't have `mallocProvider.get` and `mallocProvider.free` in C, you have `malloc` and `free`, and inside the allocated object, there's a header/footer that tracks global state.

On `MarshalWithPool(pool *Pool) ([]byte, error)`, you need to return an object that can be tracked and recycled, and `[]byte` can't have that and can't have the above safeguards, so we need to return `(*ManagedObject, error)`, which right now is `(*mem.Bytes, error)`.

@johanbrandhorst @jmarais @donaldgraham would love your input. I'm going to play with the more global solution today and see if I can eliminate the overhead of the pooling code completely, so that when you are "using" the pool, but the pool is disabled and just allocates new memory every time, there is almost no overhead.

Current benchmarks:

```
BenchmarkPoolFalseWithoutMaps-8                     	 1000000	      1824 ns/op	    1664 B/op	      33 allocs/op
BenchmarkPoolTrueDisabledWithoutMaps-8              	 1000000	      2261 ns/op	    2208 B/op	      35 allocs/op
BenchmarkPoolTrueDisabledWithSegListWithoutMaps-8   	 1000000	      2282 ns/op	    2256 B/op	      36 allocs/op
BenchmarkPoolTrueEnabledWithoutMaps-8               	  500000	      2673 ns/op	     112 B/op	       1 allocs/op
BenchmarkPoolTrueEnabledWithSegListWithoutMaps-8    	  500000	      2801 ns/op	       0 B/op	       0 allocs/op
BenchmarkPoolFalseWithMaps-8                        	  300000	      5121 ns/op	    3540 B/op	      59 allocs/op
BenchmarkPoolTrueDisabledWithMaps-8                 	  200000	      6240 ns/op	    4404 B/op	      61 allocs/op
BenchmarkPoolTrueDisabledWithSegListAndMaps-8       	  200000	      6179 ns/op	    4452 B/op	      62 allocs/op
BenchmarkPoolTrueEnabledWithMaps-8                  	  200000	      7654 ns/op	    1028 B/op	       7 allocs/op
BenchmarkPoolTrueEnabledWithSegListAndMaps-8        	  200000	      7713 ns/op	     772 B/op	       6 allocs/op
```

`FalseWithoutMaps` needs to be basically the same as `TrueDisabledWithoutMaps`, which I think I can accomplish with a clean API using global message pools generated per message type.

peter-edge:
@johanbrandhorst that looks to be a bug, but I think it will be automatically fixed with the refactor I'm about to do with the globals. Basically there won't be any type assertions at all.

protoc-gen-gogo/generator/generator.go
peter-edge: Yea I don't know, it's probably fine to just do `"0"`, but hey if we're doing the switch then why not.

peter-edge:
@johanbrandhorst I figured out how to go from `FieldDescriptorProto` to `FileDescriptorProto` so the above should be fixed

peter-edge:
Never mind, still one bug.

peter-edge:
Update @awalterschulze @johanbrandhorst code is now completely revamped and is _much_ cleaner, speed with going through all pooling logic but pooling disabled is within about 10-15% now of not going through the pooling logic at all, I want to get that number closer to 2-3%, but now that there's no more maps for messages anymore, it should be possible. Next thing to tackle is the seg list. Take a look, the surface area is much smaller now.

peter-edge:
@johanbrandhorst the above failure you had should be fixed now

johanbrandhorst:
I think you mean @jmarais?

jmarais:
@peter-edge Cool. 
With your change I see we skip well known types now.

```golang
if gogoproto.GeneratesPool(p.ObjectNamed(field.GetTypeName()).File().FileDescriptorProto) {
  p.P(`m.`, fieldName, `.Recycle()`)
}
```
the check returns `false` for the `any` type in my test.

@awalterschulze Does this option apply to all messages, or just in this file.
```
option (gogoproto.pool) = true;
```
If I import and use any:
```
import "google/protobuf/any.proto";

message Foo {
  google.protobuf.Any any = 6;
}
```
when the pool plugin does the pool option check we get false. I am not 100% sure what the interaction should be.
How does other gogo options affect the well known types?

awalterschulze:
Typically gogoproto options are set per message and then gogooptions with a `_all` suffix set all messages per file.

For now I am more concerned with the speed, but we should definitely also tackle these edge cases once the speed is sorted.  Pooled objects should be able to interact with non pooled objects, since users don't always have control over their child objects, case and point being well known types.  Good catch.

peter-edge:
@awalterschulze @jmarais this is expected behavior - the Well-Known Types files do not have the `(gogoproto.pool)` option turned on in their files. This actually demonstrates what @awalterschulze pointed out we want - that non-pooled messages can interact with pooled messages without an issue. I've explicitly verified this as well.

We probably would want to turn on the memory pooling option for the Well-Known Types, once I have the disabled speed equal to the not-generated-at-all speed. Right now it's close.

And with that, an update: the latest changes to (1) not use maps or type assertions via the globals (2) use the globals to cut short earlier on the pooling-disabled path, disabled pooling is now about 10-15% slower than no-generated-code for pooling at all, and enabling pooling is quite literally the same speed as no-generated-code at all (meaning we got a speed bump from pooling):

```
BenchmarkPoolFalseWithoutMaps-8                                 	 1000000	      1834 ns/op	    1664 B/op	      33 allocs/op
BenchmarkPoolTrueDisabledWithoutMaps-8                          	 1000000	      1988 ns/op	    2208 B/op	      35 allocs/op
BenchmarkPoolTrueDisabledWithSegListWithoutMaps-8               	 1000000	      2044 ns/op	    2256 B/op	      36 allocs/op
BenchmarkPoolTrueEnabledWithoutMapsChannelSize0-8               	 1000000	      1868 ns/op	     112 B/op	       1 allocs/op
BenchmarkPoolTrueEnabledWithSegListWithoutMapsChannelSize0-8    	 1000000	      1832 ns/op	       0 B/op	       0 allocs/op
```

I need to figure out why the disabled route is allocating more, and why it is slower. Once that is figured out, the entire codepath for memory pooling is a no-op if you don't enable it, meaning in theory we don't even need to specify a file option for whether users generate the pooling code or not, we could just generate it all the time. However, I'd feel safer generating the existing code if pooling is not explicitly asked for, so there's zero changes to existing generated code.

peter-edge:
Note the terminology:

- `PoolFalse` - the code used here does not have the file option `(gogoproto.pool)` set, so this represents the existing generated code and code path
- `PoolTrueDisabled` - this is where the pooling code is generated, and we're using `GetFoo` in the benchmarks and in the unmarshal functions, but pooling is disabled at runtime, so no pooling is actually occuring. This is the path I want to be 100% equal to the existing code, so that pooling or not is a runtime decision.
- `PoolTrueEnabled` - this is when pooling is enabled. This is where we have no allocations on the benchmark (or one allocation if we do not use `MarshalPool`, the `WithSegList` benchmarks use `MarshalPool` instead of `Marshal`). This should in theory be the fastest. Right now the speed is equal to `PoolFalse`.

peter-edge:
@awalterschulze @jmarais almost there, I removed the pool field from each message and instead put messages back into the global pool (which has the same effect, with the globals there's no need to have the reference anymore), and memory usage on the disabled path dropped from ~2200B/op to ~1800B/op, reflecting the lack of a need for each eight-byte pointer on each message. Speed has gotten very close now, still a couple kinks:

```
nchmarkPoolFalseWithoutMaps-8                                 	 1000000	      1838 ns/op	    1664 B/op	      33 allocs/op
BenchmarkPoolTrueDisabledWithoutMaps-8                          	 1000000	      1960 ns/op	    1856 B/op	      35 allocs/op
BenchmarkPoolTrueDisabledWithSegListWithoutMaps-8               	 1000000	      2056 ns/op	    1904 B/op	      36 allocs/op
BenchmarkPoolTrueEnabledWithoutMapsChannelSize0-8               	 1000000	      1857 ns/op	     112 B/op	       1 allocs/op
BenchmarkPoolTrueEnabledWithSegListWithoutMapsChannelSize0-8    	 1000000	      1859 ns/op	       0 B/op	       0 allocs/op
```

peter-edge:
Note that if I add 100 elements instead of 10 to the repeated message list within the test message, performance is a lot closer:

```
BenchmarkPoolFalseWithoutMaps-8                                 	  100000	     12327 ns/op	   14800 B/op	     219 allocs/op
BenchmarkPoolTrueDisabledWithoutMaps-8                          	  100000	     12668 ns/op	   14992 B/op	     221 allocs/op
BenchmarkPoolTrueDisabledWithSegListWithoutMaps-8               	  100000	     12847 ns/op	   15040 B/op	     222 allocs/op
BenchmarkPoolTrueEnabledWithoutMapsChannelSize0-8               	  100000	     15109 ns/op	    1028 B/op	       1 allocs/op
BenchmarkPoolTrueEnabledWithSegListWithoutMapsChannelSize0-8    	  100000	     14816 ns/op	       0 B/op	       0 allocs/op
```

peter-edge:
After a lot of investigation, I've come to the conclusion that the benchmarks are different because of a compiler optimization that doesn't reflect real-world usage.

At the bottom of `benchmarkTrue`, you'll see effectively:

```go
	unmarshaledFoo := pooltrue.GetFoo() // (1) referenced below
	//unmarshaledFoo := &pooltrue.Foo{} // (2) referenced below
	if err := unmarshaledFoo.Unmarshal(data); err != nil {
		b.Fatal(err)
	}
	unmarshaledFoo.Recycle() // (3) referenced below
```

In `benchmarkFalse`, you have:

```go
	unmarshaledFoo := &poolfalse.Foo{} // (4) referenced below
	if err := unmarshaledFoo.Unmarshal(data); err != nil {
		b.Fatal(err)
	}
```

There are similar lines for other allocations.

The `False` and `TrueDisabled` benchmarks become the same if you switch lines (1) and (2), and comment out line (3). Note that commenting out line (3) in `Disabled` should have no effect, as this is a no-op. Switching lines (1) and (2) should also have no effect other than , as (1) just returns (2). However, if EITHER (1) or (3) is there, you get an extra allocation (if you only switch (1) for (2), or only comment out (3), there's an extra allocation. If you do both, there is not an extra allocation).

If I make it so that exactly one object is being allocated with `GetFoo` in the entire benchmark, the benchmark goes up by 96B/op. Coincidentally, this is exactly the size of an empty `*Foo` if you look at the definition (2 * 8 for maps, since its a list of map entries, each of which have two 8-byte fields).

**What this means is that Golang is not making an allocation if you only access a struct inside the same function you created it. This makes sense.**

What's effectively happening is that Golang is saying `OK, I can guarantee that you are just operating on this struct within this stack, so I'm not going to allocate it on the heap. When you exit the function, there is no cleanup to do`. However, if you create it using `GetFoo`, Golang had to return the `Foo` from another function, so it went on the heap. When you do `Recycle()`, which is effectively `func RecycleFoo(foo *Foo)`...well actually if Golang was really smart, it could realize that Recycle won't go outside the stack boundary either, but that's too much to ask, it assumes at this point that `Foo` could be saved somewhere else, and it allocates it on the heap.

I'm working to reproduce this for `poolfalse` to verify that this is what is going on.

So what does this mean?

1. These benchmarks, just as they are, don't 100% represent real-world conditions, as in general, you will almost always be passing around Protobuf messages between functions, and they won't just stay on the stack. Once I've reproduced this, I hope to show that with the updates I made today, pooling is the same if you just disable it at runtime, and about 10-15% **faster** when you use pooling at this point (this is what I expect once I trick Golang into putting `poolfalse` objects from the benchmark on the heap).
2. Our documentation for pooling will note that if you know an object will be short-lived and only be used within your function, and you will not pass it to any other function (note that these conditions are very very unusual/rare for Protobuf messages), that you should _not_ use the pooling, and instead do a naked allocation like you do now using `&Foo{}`.

peter-edge:
And voila! Disabled pooling is now the same speed as no generated code for pooling at all, and enabled pooling is **5% faster now**, still with no allocations (or one allocation if you don't use `MarshalPool`).

```
BenchmarkPoolFalseWithoutMaps-8                                 	 1000000	      1854 ns/op	    1664 B/op	      33 allocs/op
BenchmarkPoolFalseWithoutMapsWithHeap-8                         	 1000000	      1916 ns/op	    1856 B/op	      35 allocs/op
BenchmarkPoolTrueDisabledWithoutMaps-8                          	 1000000	      1961 ns/op	    1856 B/op	      35 allocs/op
BenchmarkPoolTrueDisabledWithSegListWithoutMaps-8               	 1000000	      2066 ns/op	    1904 B/op	      36 allocs/op
BenchmarkPoolTrueEnabledWithoutMapsChannelSize0-8               	 1000000	      1845 ns/op	     112 B/op	       1 allocs/op
BenchmarkPoolTrueEnabledWithSegListWithoutMapsChannelSize0-8    	 1000000	      1820 ns/op	       0 B/op	       0 allocs/op
```

The `FalseWithoutMaps` is not allocating anything on the heap, the `FalseWithoutMapsWithHeap` is allocating on the heap, like most Protobuf structs will be. You'll see that the `TrueDisabledWithoutMaps` cases are almost identical to `FalseWithoutMapsWithHeap`, and the `TrueEnabledWithoutMaps` are about 5% faster.

With this, pooling is ready for prime-time, although I'm still going to work on optimizations. We will note that due to Golang's escape analysis (that some part of me knew about, but somehow just actually read the mechanics of https://www.ardanlabs.com/blog/2017/05/language-mechanics-on-escape-analysis.html), if you are exclusively using a Protobuf struct within your current stack (or child stacks), it is recommended to do naked allocation with `&Foo{}` instead of `GetFoo()` as no memory will be created, and Golang can take advantage of this by not allocating anything on the heap.

@awalterschulze @jmarais @johanbrandhorst

peter-edge:
I've put up integration into the gRPC plugin for generated unary methods (streaming is just as trivial, but just for minimal work while I get feedback). The general concept is that the generated servers always create request messages and then pass them on to the handler, so why not make them pooled before passing them onto the server? It's a minimal diff in the grpc plugin code:

```
diff --git a/protoc-gen-gogo/grpc/grpc.go b/protoc-gen-gogo/grpc/grpc.go
index 6a33ec71..f58d26d7 100644
--- a/protoc-gen-gogo/grpc/grpc.go
+++ b/protoc-gen-gogo/grpc/grpc.go
@@ -40,6 +40,7 @@ import (
        "strconv"
        "strings"

+       "github.com/gogo/protobuf/gogoproto"
        pb "github.com/gogo/protobuf/protoc-gen-gogo/descriptor"
        "github.com/gogo/protobuf/protoc-gen-gogo/generator"
 )
@@ -408,7 +409,17 @@ func (g *grpc) generateServerMethod(servName, fullServName string, method *pb.Me

        if !method.GetServerStreaming() && !method.GetClientStreaming() {
                g.P("func ", hname, "(srv interface{}, ctx ", contextPkg, ".Context, dec func(interface{}) error, interceptor ", grpcPkg, ".UnaryServerInterceptor) (interface{}, error) {")
-               g.P("in := new(", inType, ")")
+
+               inDesc := g.gen.ObjectNamed(method.GetInputType())
+               inPkgname := g.gen.DefaultPackageName(inDesc)
+               inBasename := generator.CamelCaseSlice(inDesc.TypeName())
+               if gogoproto.HasPool(inDesc.File().FileDescriptorProto) {
+                       g.P(`in := `, inPkgname, `Get`, inBasename, `()`)
+                       g.P(`defer in.Recycle()`)
+               } else {
+                       g.P("in := new(", inType, ")")
+               }
```

And results in a minimal diff in the generated code:

```
diff --git a/test/pool/pooltrue/pooltrue.pb.go b/test/pool/pooltrue/pooltrue.pb.go
index 40a5e322..89ecf490 100644
--- a/test/pool/pooltrue/pooltrue.pb.go
+++ b/test/pool/pooltrue/pooltrue.pb.go
@@ -223,7 +223,8 @@ func RegisterBazServer(s *grpc.Server, srv BazServer) {
 }

 func _Baz_Hello_Handler(srv interface{}, ctx context.Context, dec func(interface{}) error, interceptor grpc.UnaryServerInterceptor) (interface{}, error) {
-       in := new(Foo)
+       in := GetFoo()
+       defer in.Recycle()
        if err := dec(in); err != nil {
                return nil, err
        }
```

**What is on this PR now will not be the final implementation. Whether or not to do this should be a method option.**. Users may store references to requests and keep them around after the function returns, and if they do this and we recycle the message, they are in trouble. This should be an opt-in method option that says if the request message has pooling enabled, to do the pooling automatically.

Putting this up for feedback.

johanbrandhorst:
Just so I understand correctly; what you're saying is that the use of `pooled` messages is "contagious"? Anyone importing a pooled message/file anywhere in their stack may as well sign up to pooling because they already have to be careful with the use of `Recycle`? We'll have to be super explicit about this.

Personally, I think I might still want the power of using it per-message. I think there's a usecase for users who want to opt into pooling only for part of their stack, and if I understand correctly, as long as it's an isolated subgraph it doesn't have to affect the rest of the messages, right?

peter-edge:
If someone imports a pooled message and they don't themselves enable pooling, there's no change. Pooling or not, regardless of whether you generate the code, is a runtime decision through `mem.EnablePooling()`. So if someone generates a file with `(gogoproto.pool) = true`, and someone else (who is not aware of pooling) imports it, there's no effect.

Just for clarification: regardless of whether you turn pooling on in the generation, you have the power to use it per-message. The option only tells you whether or not you generate the functionality to have pooling available, not whether or not you use it on a given call.

Assuming you've called `mem.EnablePooling()` and you have a message `pooled.Foo`:

- To use it unpooled, do `foo := &pooled.Foo{}` like you normally do.
- To use it pooled, do `foo := pooled.GetFoo()`. Then make sure to call `foo.Recycle()`, probably in a `defer` call on the next line.

That's it.

Two changes I'm probably going to make:

- I'm probably going to rename `GetFoo` to `GetPooledFoo` to make it explicit what is happening. Again, if you don't call `mem.EnablePooling()`, then `GetPooledFoo` is the same as doing `&Foo{}`, except it will always get allocated on the heap (see the discussion above).
- I need to add a marker to each message when doing pooling that contains two pieces of information: whether or not the message was allocated from a pool, and whether or not the message is recycled. I want to store whether or not the message was allocated as a safety mechanism so that if users call `Recycle()` but are not allocating messages using the pooling, we don't just fill up the pool blindly (which they shouldn't do, but this could be a common bug, and I want `Recycle` to be defined as a no-op if you did not allocate the message from a pooling). I want the marker of whether or not a message has been recycled but not reset so that we can panic if a user uses any method (Marshal/Unmarshal/Getters/etc) on the message after a message has been recycled. Using a message after recycle is the equivalent of a nil pointer exception, except without this safeguard and will be a lot harder to debug.

I can do the marker in a `uint8` and only use one byte per message (another reason we want to opt-in to pool code generation), but we might want to use a uint32 for alignment (whether or not a message is word-aligned is can vary depending on the set fields, but if we have a uint8, it will almost for sure not be word-aligned).

Does that address the question?

johanbrandhorst:
Yes, thanks, that's great.

So the killer benchmark would be if we could load up a gRPC server with and without pooling and show a significant difference in CPU/latency/allocations with this. Any ideas if that's achievable? I understand it'd probably go in it's own repo.

johanbrandhorst:
I'd love to help any way I can if course :).

peter-edge:
@johanbrandhorst there will probably be API changes, but the basic functionality is now there for grpc request/response pooling. If you could benchmark something real that would be amazing.

See `test/pool/pooltrue/pooltrue.proto` for an example of how it works:

```proto
service Baz {
  rpc Hello(Foo) returns (Bar) {
    option (gogoproto.create_pooled_server_requests) = true;
    option (gogoproto.create_pooled_client_responses) = true;
  }
  rpc HelloClientStream(stream Foo) returns (Bar) {
    option (gogoproto.create_pooled_server_requests) = true;
    option (gogoproto.create_pooled_client_responses) = true;
  }
  rpc HelloServerStream(Foo) returns (stream Bar) {
    option (gogoproto.create_pooled_server_requests) = true;
    option (gogoproto.create_pooled_client_responses) = true;
  }
  rpc HelloBidiStream(stream Foo) returns (stream Bar) {
    option (gogoproto.create_pooled_server_requests) = true;
    option (gogoproto.create_pooled_client_responses) = true;
  }
}
```

Let's do over what this does:

For the server:

- For a unary method or server streaming method, server request `Foo`s will be created from the pool and recycled automatically. The user doesn't have to do anything. This should be enabled if the user does not store a reference to the request after returning from the handler.
- For a client streaming or bidi streaming method, server requests will be created from the pool and returned from `Recv`, and the caller of `Recv` is responsible for calling `Recycle` on the returned `Foo`.

For the client:

- For all methods, client response `Bars` will be created from the pool. The caller is responsible for calling `Recycle` on the returned `Bar`.

See the generated code for more details. Finding a good project to test this against would be amazing. You'll have to add the method options to the appropriate methods, and make do any necessary edits to the implementation in terms of calling `Recycle`. To start, the easiest would be to just enable `(gogoproto.create_pooled_server_requests)` for unary and server streaming methods, since `Recycle` is called automatically so there is no code change needed.

johanbrandhorst:
@peter-edge This is some amazing stuff, unfortunately I've really no idea where to even start on benchmarking gRPC server performance. I think there's various stress testing libraries out there but I've very little experience with any of them. I'll let you know if I get around to doing something :).

peter-edge:
It turns out etcd has a benchmark tool https://github.com/coreos/etcd/tree/master/tools/benchmark so this makes testing it easier. I setup etcd to use my fork and the branch this is on, and with a few other changes got it to build against this. I also discovered a few bugs, noting here so I can fix them:

- Fields that have `nullable = false` are not properly handled. The `Reset` method incorrectly sets them to `nil`, and the `Recycle` method incorrectly recycles them.
- Oneofs are not properly handled. This will be more involved, I just completely missed it.

johanbrandhorst:
@peter-edge Amazing! Looking forward to seeing the results!

awalterschulze:
I am also very interested to see the real world benchmark

PR #463, opened 2018-08-14
Remove XXX_ fields from bson for timestamp type

JulienBreux:
Hi,

I create this PR just because we store directly the Timestamp struct to Mongdb.

And now, since the `XXX_` fields creations.
We have not just 2 entries but 5 entries.

And it's not cool to convert/marshal/unmarshal/compute/and/more just because `XXX_` private fields are now.

Do you have an opinion about that?


AlmogBaku:
A. you should use the `gogofaster` compiler, also you can add custom tags
B. never manually edit autogenerated code

Issue #411, opened 2018-06-01
MarshalJSONPB is not called for custom type if it's an alias to a primitive


yurishkuro:
I am using a custom type

```go
type SpanID uint64

// MarshalJSONPB renders span id as a single hex string.
func (s SpanID) MarshalJSONPB(*jsonpb.Marshaler) ([]byte, error) {
	// something
}
```

When serializing to JSON using `github.com/gogo/protobuf/jsonpb` I realized that its MarshalJSONPB method is never called, even though it works for other custom types that are defined as structs, instead of a type alias. Instead, the stdlib's `json` package is used:
```
> github.com/jaegertracing/jaeger/vendor/github.com/gogo/protobuf/jsonpb.(*Marshaler).marshalValue() ./vendor/github.com/gogo/protobuf/jsonpb/jsonpb.go:682 (PC: 0x16927a9)
   677:				return out.err
   678:			}
   679:		}
   680:
   681:		// Default handling defers to the encoding/json library.
=> 682:		b, err := json.Marshal(v.Interface())
   683:		if err != nil {
```

which happens to call `MarshalText()`.

awalterschulze:
Have you tried https://github.com/gogo/protobuf/blob/master/custom_types.md
I think you need to `func (t T) MarshalJSON() ([]byte, error) {}` and not `MarshalJSONPB` if that makes sense?

yurishkuro:
Hm, I had those before but ran into some other issue. Notably, if the custom type implements `jsonpb.JSONPBUnmarshaler`, then `UnmarshalJSONPB` is called on it even if it defines `UnmarshalJSON`.

Thanks.

yurishkuro:
Actually, I think there's still something wrong. I changed the code to only provide MarshalJSON / UnmarshalJSON functions, and my parsing tests started breaking, now on the parsing side. When I restore `UnmarshalJSONPB` function, I get successful run:

```
$ go test -v -run TestSpanIDUnmarshalJSON ./model
=== RUN   TestSpanIDUnmarshalJSON
=== RUN   TestSpanIDUnmarshalJSON/{"traceID":"0","spanID":"1"}
SpanID.UnmarshalJSONPB("1") called
=== RUN   TestSpanIDUnmarshalJSON/{"traceID":"0","spanID":"f"}
SpanID.UnmarshalJSONPB("f") called
=== RUN   TestSpanIDUnmarshalJSON/{"traceID":"0","spanID":"1f"}
[...]

PR #470, opened 2018-08-22
added test for MarshalJSONPB not being called for custom type if it's…

donaldgraham:
… an alias to a primitive (#411)


donaldgraham:
We now need to resolve the failing test from @yurishkuro.

PR #483, opened 2018-09-12
Adding missing func to CustomType documentation

AlmogBaku:
The function `Size()` is mandatory for `CustomType` and should be documented


jmarais:
Thanks.

PR #486, opened 2018-09-20
Add goproto_morexxx and goproto_morexxx_all

virtuald:
- Allows omitting XXX fields when using moretags

There have been a variety of proposals to allow changing struct tags (#368 in particular).. but we already have moretags, and it's Good Enough for what I want now -- except that when I serialize I get those stupid XXX tags in my output. I'd like them to go away, so here's my solution for that.

If there's a vanity binary solution to this I'm also open to that, but I didn't see an easy way to make it work.


dehypnosis:
Wait for this feature!!

virtuald:
Rebased, fixed protobufs.

PR #498, opened 2018-10-14
Fix text unmarshal for (u)int(8/16) fields

aspcartman:
Consider the following proto:
```
message Message {
    uint32 field = 1 [(gogoproto.casttype) = "uint8"];
}
```

A completely valid proto with a valid `gogoproto.casttype`. 
But the generated tests will fail because of the inability of the text parser to parse value into (u)int(8/16). This PR fixes the issue.


donaldgraham:
This is great.  Thanks. 
 
Would you mind adding a test for this?

aspcartman:
@donaldgraham of course : ) Added as test/issue498 and verified: test fails on master branch but passes after the fix.

aspcartman:
@donaldgraham of course : ) Added as test/issue498 and verified: test fails on master branch but passes after the fix.

aspcartman:
@donaldgraham of course : ) Added as test/issue498 and verified: test fails on master branch but passes after the fix.

jmarais:
Thanks for the test and the contribution. Looks good.

PR #529, opened 2018-12-13
support clone and merge of non-nullable fields

RaduBerinde:
Cloning or merging is broken if a message uses a repeated non-nullable
struct field (segmentation fault). This change implements the handling
of a non-pointer slice and adds clone and merge tests for non-nullable
fields.

@awalterschulze @jmarais @donaldgraham

---
This change is [Reviewable](https://reviewable.io/reviews/gogo/protobuf/529)

### DIFF
--- a/proto/table_merge.go
+++ b/proto/table_merge.go
@@ -530,6 +530,27 @@ func (mi *mergeInfo) computeMergeInfo() {
 			}
 		case reflect.Struct:
 			switch {
+			case !isPointer && isSlice:
+				// This case is for the gogoproto nullable extension, specifically when
+				// we have a repeated struct with nullable=false.
+				mergeInfo := getMergeInfo(tf)
+				mfi.merge = func(dst, src pointer) {
+					dstSlice := dst.getSlice(tf)
+					srcSlice := src.getSlice(tf)
+
+					numExisting := dstSlice.Len()
+					numNew := srcSlice.Len()
+					newSlice := reflect.MakeSlice(reflect.SliceOf(tf), numExisting+numNew, numExisting+numNew)
+					reflect.Copy(newSlice, dstSlice)
+					// Clone each element from src.
+					for i := 0; i < numNew; i++ {
+						mergeInfo.merge(
+							valToPointer(newSlice.Index(numExisting+i).Addr()),
+							valToPointer(srcSlice.Index(i).Addr()),
+						)
+					}
+					dstSlice.Set(newSlice)
+				}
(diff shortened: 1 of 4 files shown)


RaduBerinde:
Please let me know if there is anything I can do to get this merged soon (it's needed for CockroachDB to update its proto deps).

RaduBerinde:
> Why are some tests commented out?

Oops, I meant to remove the old tests (they are very basic and the new tests cover everything they cover).

I'll look over the testgen stuff, looks interesting.

RaduBerinde:
Unfortunately with testgen, we find all sorts of unimplemented cases (e.g. "merger not found for type:custom.Uint128"). Do you advise trying to fix all of these in this PR?

RaduBerinde:
After looking further, I think the problems that need addressing are not easy to fix. In particular, custom types will need to define a method for merging data, and there's a design discussion to be had there.

I am not sure how to proceed with the testgen stuff in this PR. I can think of two possibilities:
 - disable the generated clone test unless a special build tag is defined; we can add a (temporary) target that runs the tests with that tag
 - add `testgen_no_clone` / `testgen_no_clone_all` extensions and use them for the tests that fail, with TODOs to fix the underlying issue remove them

dsnet:
To fix this issue, it seems that converging with the [v2 implementation of `golang/protobuf`](https://github.com/golang/protobuf/tree/api-v2) is the right answer. The v2 API is all about making [protobuf reflection a first-class feature](https://goo.gl/96gGnE). With reflection, the way to implement merge becomes agnostic towards the actual underlying implementation (whether it be golang/protobuf, gogo/protobuf, or some other custom type). The requirement is that the type know how to provide reflection information about itself.

RaduBerinde:
Is there any pathway to get this particular fix merged without boiling the ocean? If not I will have to fork the repo.

awalterschulze:
Ok so fixing custom type would require more work.

But your fix before my testgen proposal already improves the current state
of clone and merge.

I look forward to the apiv2 world and I am a little out of the loop, but I
assume there is still a lot of work left.

So maybe your original commit with some minor clean up is the pragmatic
route at this time, but this up to @donaldgraham and @jmarais the new
maintainers of gogoprotobuf

knz:
I have tested this in v1.3.1 and I confirm the bug was fixed there. Upstream patch here: 

https://github.com/golang/protobuf/commit/8cc9e46429bfb16289d40d30b2ee3f4923b47345

PR #552, opened 2019-03-06
Add space in markdown header

TennyZhuang:


donaldgraham:
Thanks for picking this up and fixing it.

PR #567, opened 2019-05-05
support buptbill220/protobuf gogofmqjson

buptbill220:
Change-Id: Iaab4149b12991e00d2e47e30e7c7a0476ceb82f9


jmarais:
Hi. It seems like there are a lot of unnecessary changes with your diff. 
What feature did you try to add with this pull request?

buptbill220:
> Hi. It seems like there are a lot of unnecessary changes with your diff.
> What feature did you try to add with this pull request?

just suply in some case like this: I have a message, maybe use json or protobuf or anything else protocol. We just describe its protocol with protobuf, but encoding or decoding will be transparent for developer. Developer just tell us its protocol and codec type, then we generate coder file. 
exampler:
type Codec interface {
	Marshal(msg interface{}) ([]byte, error)
	Unmarshal (data []byte, msg interface{}) (error)
}

User.proto
message User {
    required int64 id = 1;
    required string name = 2
    required int32 age = 3;
    optional Occupation occ = 4;
    repeated Occupation occs = 5;
}
message Occupation {
    required int64 id = 1;
    required string industry = 2;
}


user := &User{}
user.Unmarshal & user.Marshal, don't case json or protobuf.

PR #576, opened 2019-06-07
Add optional support for FieldMask wkt

gtungatkar:
Go implementation of JSONPB does not support well-known proto type FieldMask, which has special JSON encoding. 
The spec is defined as:
`
mask {
  paths: "user.display_name"
  paths: "photo"
}
`
JSON:
`
{
  mask: "user.displayName,photo"
}
`
What jsonpb actually serializes it as:
` mask: {paths: ["user.displayName","photo"]}`

While this has been fixed in v2 as described in this issue https://github.com/golang/protobuf/issues/745 , v1 is still widely in use and v2 does not have published timelines.

**What does this pull request do**
This PR adds support for correctly serializing FieldMask. But since this is a breaking change, this fix is gated by an env variable. 
Protobuf implementations in other languages like python/cpp already support FieldMask. This allows gogo JSONPB generated JSON to be compatible.


gtungatkar:
@donaldgraham @jmarais please let me know if I can do anything to get this reviewed and hopefully merged.

AlexAPB:
Ping on this PR! We'd like to get this fixed :)

jsonpb/jsonpb_test_proto/field_mask_test_object.proto
```diff
@@ -0,0 +1,40 @@
...
+
+syntax = "proto3";
+
+import "google/protobuf/field_mask.proto";
+
+package jsonpb;
+
+message FieldMaskWKT {
+  google.protobuf.FieldMask field_mask = 1;
+}
```
AlexAPB:
```suggestion
}

```

jsonpb/jsonpb_test.go
```diff
@@ -520,9 +522,11 @@ var marshalingTests = []struct {
 	{"BytesValue", marshaler, &pb.KnownTypes{Bytes: &types.BytesValue{Value: []byte("wow")}}, `{"bytes":"d293"}`},
 	{"required", marshaler, &pb.MsgWithRequired{Str: proto.String("hello")}, `{"str":"hello"}`},
 	{"required bytes", marshaler, &pb.MsgWithRequiredBytes{Byts: []byte{}}, `{"byts":""}`},
+	{"FieldMask", marshaler, &pb.FieldMaskWKT{FieldMask: &types.FieldMask{Paths: []string{"a", "b", "c_d"}}}, `{"fieldMask":"a,b,cD"}`},
```
AlexAPB: Curious, is the behaviour of `c_d` -> `cD` documented anywhere? I couldn't find anything.

jsonpb/jsonpb.go
```diff
@@ -57,6 +58,27 @@ import (
...
+//}
+// Current library does not correctly serialize FieldMask, and generates following JSON
+//{
+//  mask: {paths: ["user.displayName","photo"]}
+//}
+// SUPPORT_FIELD_MASK_WKT_JSON is an environment variable to allow clients to opt in to the fix to correctly serialize FieldMask
+// if SUPPORT_FIELD_MASK_WKT_JSON is set to 1, FieldMask will be handled correctly and JSON generated will match spec
+// If undefined or unset, no change to FieldMask to JSON conversion, so that any existing users of the library do not break.
+// Fix to correctly serialize FieldMask is taken from https://github.com/golang/protobuf/blob/api-v2/encoding/protojson/well_known_types.go
+var supportFieldMaskWKT = os.Getenv("SUPPORT_FIELD_MASK_WKT_JSON") == "1"
```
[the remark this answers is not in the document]
gtungatkar: Internally, we can drop the flag. But, as a patch to gogo protobuf, this is a breaking change.

jsonpb/jsonpb.go
```diff
@@ -959,6 +1001,29 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
 			return nil
 		}
 	}
+	//Clients want to support FieldMask WKT correctly
+	//Largely follows the implementation upstream in golang/protobuf, with some changes to conform to gogo
+	//https://github.com/golang/protobuf/blob/a3369c5dc28ac197f8ec8de40a7291be4dea85b5/encoding/protojson/well_known_types.go#L903
+	//unmarshalFieldMask
+	if supportFieldMaskWKT {
+		type xname interface {
```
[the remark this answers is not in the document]
gtungatkar: It is not used anywhere else, so i wanted to keep it close to where this patch is using it.

jsonpb/jsonpb_test.go
gtungatkar: Check the display_name example here https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/field-mask

jsonpb/jsonpb.go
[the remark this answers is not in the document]
gtungatkar: Removed the flag so that it is easy to apply the patch internally.

jsonpb/jsonpb_test.go
AlexAPB: thanks!

jsonpb/jsonpb_test.go
```diff
@@ -860,9 +880,11 @@ var unmarshalingTests = []struct {
 	{"null BytesValue", Unmarshaler{}, `{"bytes":null}`, &pb.KnownTypes{Bytes: nil}},
 	{"required", Unmarshaler{}, `{"str":"hello"}`, &pb.MsgWithRequired{Str: proto.String("hello")}},
 	{"required bytes", Unmarshaler{}, `{"byts": []}`, &pb.MsgWithRequiredBytes{Byts: []byte{}}},
+	{"FieldMask", Unmarshaler{}, `{"fieldMask":"a,b,cD"}`, &pb.FieldMaskWKT{FieldMask: &types.FieldMask{Paths: []string{"a", "b", "c_d"}}}},
 }
 
 func TestUnmarshaling(t *testing.T) {
+
```
mh-park: Odd newline

jsonpb/jsonpb.go
```diff
@@ -264,7 +264,26 @@ func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeU
...
+	// https://github.com/golang/protobuf/blob/a3369c5dc28ac197f8ec8de40a7291be4dea85b5/encoding/protojson/well_known_types.go#L884
+	if proto.MessageName(v) == "google.protobuf.FieldMask" {
+		list := s.Field(0)
+		paths := make([]string, 0, list.Len())
+		for i := 0; i < list.Len(); i++ {
+			s := list.Index(i).String()
+			// Return error if conversion to camelCase is not reversible.
+			cc := JSONCamelCase(s)
+			if s != JSONSnakeCase(cc) {
+				return fmt.Errorf("paths contains irreversible value %q", s)
```
mh-park: Can we collect the errors here with `multierr`? With the current error logic, it would cut at the first error, and at a user's perspective, if they have multiple errors, they would have to run the compiler, fix the error, and then run the compiler again to see the next error.

jsonpb/jsonpb.go
```diff
@@ -959,6 +978,27 @@ func (u *Unmarshaler) unmarshalValue(target reflect.Value, inputValue json.RawMe
...
 	}
+	// To correctly support unmarshaling field masks, the following is largely copied from the `golang/protobuf` v2 implementation
+	//https://github.com/golang/protobuf/blob/a3369c5dc28ac197f8ec8de40a7291be4dea85b5/encoding/protojson/well_known_types.go#L903
+	//unmarshalFieldMask
+	type xname interface {
+		XXX_MessageName() string
+	}
+	if m, ok := target.Addr().Interface().(xname); ok && m.XXX_MessageName() == "google.protobuf.FieldMask" {
+		str := strings.TrimSpace(string(inputValue))
+		// Unqote the input value, similar to other WKTs (eg "Timestamp")
```
mh-park: nit: misspelling `Unqote -> Unquote`

jsonpb/jsonpb.go
gtungatkar: Keeping the error handling consistent with rest of the jsonpb and with the golang v2 implementation. multierr is not used.

jsonpb/jsonpb.go
```diff
@@ -264,7 +264,26 @@ func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeU
...
+			if s != JSONSnakeCase(cc) {
+				return fmt.Errorf("paths contains irreversible value %q", s)
+			}
+			paths = append(paths, cc)
+		}
+		// explicitly output quotation marks, similar to other WKTs (eg "Timestamp")
+		out.write(`"`)
+		out.write(strings.Join(paths, ","))
+		out.write(`"`)
+		return nil
```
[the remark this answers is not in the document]
gtungatkar: Good catch, fixed!

jmarais:
Hi. Thanks for contribution.
You mentioned:

> But since this is a breaking change, this fix is gated by an env variable.

What did you remove it in a recent commit?

gtungatkar:
> Hi. Thanks for contribution.
> You mentioned:
> 
> > But since this is a breaking change, this fix is gated by an env variable.
> 
> Why did you remove it in a recent commit? Since this is a breaking change I do not just want to merge it without any guard in place.

Hi, yes i removed it since we were just going to apply the patch internally if we could not get it accepted. I can add it back if you think that you will be able to accept/merge the pull request. I will address other comments.

jmarais:
I think this change is fine. As you pointed out, the current marshaled format is not inline with the proto spec. 
However, think a better approach would be to add it as an option on the marshaler/unmarshaler struct first, with the default off. This will prevent us from break backward compatibility.

jsonpb/jsonpb.go
```diff
@@ -264,7 +264,26 @@ func (m *Marshaler) marshalObject(out *errWriter, v proto.Message, indent, typeU
 			return m.marshalValue(out, &proto.Properties{}, x, indent)
 		}
 	}
-
+	// To correctly support marshaling field masks, the following is largely copied from the `golang/protobuf` v2 implementation
+	// https://github.com/golang/protobuf/blob/a3369c5dc28ac197f8ec8de40a7291be4dea85b5/encoding/protojson/well_known_types.go#L884
+	if proto.MessageName(v) == "google.protobuf.FieldMask" {
```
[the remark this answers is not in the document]
gtungatkar: Done

gtungatkar:
> I think this change is fine. As you pointed out, the current marshaled format is not inline with the proto spec.
> However, think a better approach would be to add it as an option on the marshaler/unmarshaler struct first, with the default off. This will prevent us from break backward compatibility.

Added option on the marshaler/unmarshaler struct

gpopovic:
Any news about this one? :(

Issue #587, opened 2019-07-25
uint32Reader recreating buffers of same length


SebiSujar:
Hello everyone! we are profiling our app that uses an `uint32reader`. We noticed that in the ReadMsg method [here](https://github.com/gogo/protobuf/blob/master/io/uint32.go#L123) the buffer is recreated when the length is the same as the one required. This causes a buffer allocation each time a message is read with the same size as the previous.

This is not the case with the `varintReader` [here](https://github.com/gogo/protobuf/blob/master/io/varint.go#L118).

PR #588, opened 2019-07-25
Remove unnecessary alllocation of buffer in uint32Reader

SebiSujar:
Fixes issue #587


jmarais:
Thanks for the contribution.

Issue #611, opened 2019-08-29
custom tags are not generated for oneof filed


krhubert:
Hi,
 
following message:

```proto
syntax = "proto3";
  
import "gogo/protobuf/gogoproto/gogo.proto";

message T {
  oneof v {
    string s = 1 [
      (gogoproto.moretags) = 'customtag:"s"'
    ];
  }
}
```
will generate
```bash
$ protoc --gogofaster_out=. t.proto
```
go struct
```go
type T_S struct {
  S string `protobuf:"bytes,1,opt,name=s,proto3,oneof"`
}
```

So the tags are not injected into `oneof` fields.

jmarais:
Thanks for logging the bug. Please see the pull request comments.

PR #612, opened 2019-08-30
Generate custom tags for oneof fileds (#611)

krhubert:


protoc-gen-gogo/generator/generator.go
```diff
@@ -3366,6 +3364,9 @@ func (g *Generator) generateMessage(message *Descriptor) {
 
 			oneofField := oFields[*field.OneofIndex]
 			tag = "protobuf:" + g.goTag(message, field, wiretype)
+			if gogoMoreTags != nil {
+				tag = " " + *gogoMoreTags
```
jmarais: I think we needed a ```+=``` here.

jmarais:
Hi. Thanks for the pullrequest. It seems like the oneof tags were collateral damage of a large generator refactor.

protoc-gen-gogo/generator/generator.go
```diff
@@ -3366,6 +3364,9 @@ func (g *Generator) generateMessage(message *Descriptor) {
 
 			oneofField := oFields[*field.OneofIndex]
 			tag = "protobuf:" + g.goTag(message, field, wiretype)
+			if gogoMoreTags != nil {
```
jmarais: I think we can just remove any tag logic here. 
I see the tags for every message does get constructed here:
https://github.com/gogo/protobuf/blob/23325ce401c6d7afd7008f26ff87c20074ea1413/protoc-gen-gogo/generator/generator.go#L2838-L2852

We just need to use that ```tag``` for our oneofSubField struct.

jmarais:
Can we please also get a test for this issue?

I think it would make sense to add a simple oneof message under"
```test/tags/tags.proto```
And then use a xml custom tag with a marshal/unmarshal like the other tests to validate that the correct ```customtag``` has been generated.

krhubert:
@jmarais I think it's ready to review.

In tests I skip the Marshal for json and XML because I will have to implement custom marshaler for oneof interface.

krhubert:
About comments for tests look on the answer below.

> Do you have a suggestion for a test in order to validate a custom moretags field?

What do you think about using reflection and get a tag, instead of marshal/unmarshal with json/xml?

> With regards to the travis job. Remember to make in the root of the project in order to regenerate all the protos with the generator change.

Yes, I miss that. Also I found something interesting - in tags.proto syntax is set to [proto2](https://github.com/gogo/protobuf/blob/master/test/tags/tags.proto#L29), but the generated `test.pb.go` declares [GoGoProtoPackageIsVersion3](https://github.com/gogo/protobuf/blob/master/test/tags/tags.pb.go#L22) - why it so?

jmarais:
> What do you think about using reflection and get a tag, instead of marshal/unmarshal with json/xml?
That is a good suggestion. I think that will make for a nice test.

> but the generated test.pb.go declares GoGoProtoPackageIsVersion3 - why it so?
That is just a constant generated by the ```protoc-gen-gogo``` plugin when generating the ```*.pb.go``` files. The constant's definition is in ```proto``` package. It gives us a compile time check between the generated code and the current ```proto``` package.

krhubert:
@jmarais I regenerated proto definitions plus replace the test to make use of reflection.

PR #618, opened 2019-09-04
Use ProtoSize() in MarshalTo when enabled

### LINKED ISSUE #617: MarshalTo does not check ProtoSizer option in 1.3.0
Ran into a case today where when `gogoproto.protosizer_all` is enabled that bad code is generated in `MarshalTo`. This was tested against 1d64e234ea8e4bc60200c52a8f7c568bf26d1ec0

The generated code ends up looking like this:
```go
func (m *Alarm_ErrorDetails_) MarshalTo(dAtA []byte) (int, error) {
  return m.MarshalToSizedBuffer(dAtA[:m.Size()])
}
```

I've tracked it down to this line: https://github.com/gogo/protobuf/commit/65acae22fc9d1fe290b33faa2bd64cdc20a463a0#diff-a11661c9ad014a306a0799941e492826R1013 which is hard coded to use `Size()` instead of checking which to use.

Based on other code I would expect it to generate something like this:
```go
func (m *Alarm_ErrorDetails_) MarshalTo(dAtA []byte) (int, error) {
  size := m.ProtoSize()
  return m.MarshalToSizedBuffer(dAtA[:size])
}
```

gaffneyc:
When generating `MarshalTo` it was not checking the `ProtoSizer` option. I've based this off of the other places that the option is used.

Fixes #617


jmarais:
Thanks for the fix. I think you just need to call ```make``` in the project root to regenerate all the proto files in the project and then travis will pass.

gaffneyc:
Thanks, I've regenerated them and just force pushed the changes.

PR #621, opened 2019-09-08
Issue619

jmarais:
Compensate the buffer's slice when marshaling into a larger backing buf.


proto/encode.go
```diff
@@ -189,8 +189,22 @@ type Marshaler interface {
 // prefixed by a varint-encoded length.
 func (p *Buffer) EncodeMessage(pb Message) error {
 	siz := Size(pb)
+	sizVar := SizeVarint(uint64(siz))
```
apelisse: Shouldn't we be fixing `p.Marshal` rather than the calling code? If we have to change code that calls `Marshal`, then I suspect something must be wrong?

test/protobuffer/protobuffer_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+func marshalCheck(t *testing.T, m proto.Message, size int) {
+	buf := proto.NewBuffer(make([]byte, 0, size))
+	err := buf.Marshal(m)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = buf.Marshal(m)
```
apelisse: is buf.Marshal supposed to concatenate into the buffer? I didn't realize that was the expected behavior

test/protobuffer/protobuffer_test.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	bufferBytes := buf.Bytes()
+	protoBytes, _ := proto.Marshal(m)
```
apelisse: You might want to check the ignored err here

test/protobuffer/protobuffer_test.go
```diff
@@ -0,0 +1,75 @@
...
+		{"MarshalerSmall", 0, mer},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			marshalCheck(t, tc.m, tc.size)
+		})
+	}
+}
+
+func marshalCheck(t *testing.T, m proto.Message, size int) {
```
apelisse: I think I liked the test in the first version of the test (the one in the parent commit)

test/protobuffer/protobuffer_test.go
jmarais: I think this is the case. Using ```golang/protobuf``` in this same test does yield a buffer with the same message concatenated into it.

The proto/lib.go does have an explicit ```Reset()``` method. Which a client can call when he is done.
https://github.com/golang/protobuf/blob/master/proto/lib.go#L362

proto/encode.go
jmarais: I think you are right. The large change here isn't necessary

test/protobuffer/protobuffer_test.go
jmarais: Which one did you reference here? I think having a test for encoding twice into the buffer is useful. However, I can make an explicit test for twice encoding.

jmarais:
In favor of: https://github.com/gogo/protobuf/pull/627

Issue #630, opened 2019-10-01
gostring generates pointers instead of values for repeated, non-nullable fields


godfried:
example:
```proto
syntax = "proto2";
package repeatednonnullable;
import "github.com/gogo/protobuf/gogoproto/gogo.proto";
option (gogoproto.gostring_all)            = true;
message Foo {
  repeated Bar Bar1 = 1 [(gogoproto.nullable) = false];
}
message Bar {
  optional string Baz = 1 [(gogoproto.nullable) = false];
}
```
generates:
```go
// omitted for brevity
if this.Bar1 != nil {
	vs := make([]*Bar, len(this.Bar1)) // slice of pointers should be a slice of values
	for i := range vs {
		vs[i] = &this.Bar1[i]  // should just be vs[i] = this.Bar1[i] 
	}
	s = append(s, "Bar1: "+fmt.Sprintf("%#v", vs)+",\n")
}
// omitted for brevity
```
expected:
```go
// omitted for brevity
if this.Bar1 != nil {
	vs := make([]Bar, len(this.Bar1))
	for i := range vs {
		vs[i] = this.Bar1[i]
	}
	s = append(s, "Bar1: "+fmt.Sprintf("%#v", vs)+",\n")
}
// omitted for brevity
```

PR #631, opened 2019-10-01
GoString plugin: generate values instead of pointers when a field is repeated and non-nullable.

godfried:
See https://github.com/gogo/protobuf/issues/630


godfried:
I haven't regenerated any of the other protos yet.

jmarais:
Thanks for the pull request. 

Currently there is only one test case for a repeated non-nullable message, can you please add another test for the repeated nullable case?

The travis check is currently failing because of the changes in the generated pb.go files. You will need to regenerate all the pb.go files.

PR #655, opened 2020-01-06
proto: optimize global lock

TennyZhuang:
This PR is also submitted to https://github.com/golang/protobuf/pull/1004

Signed-off-by: TennyZhuang <zty0826@gmail.com>

Thie PR Use `RWMutex` to optimize `getMarshalInfo` and `getUnmarshalInfo`, for these functions, only n (number of message type) will hit Write, and m(number of message) - n will hit Read, it's the best case to use RWMutex instead of Mutex.

This optimization introduce huge improvement in our scenario.

We have 1000 worker and 1 controller, and work and controller keep heartbeat with gRPC. They also exchange the job info with each other.

The message is like

```proto
message Job {
    uint64 job_id;
    // Some other info
}

message Heartbeat {
    repeated Job jobs = 1;
}
```

About 10000 jobs in every Heartbeat, and the heartbeat QPS in controller is about 1000.

The controller handle the Heartbeat in about 10ms, and the network latency is about 10ms, but the client will use about 30s in maximum to finish a RPC call.

We use golang pprof block profile, and it seems that almost all block is caused by one global Mutex in protobuf package.

image

After optimization, in our use case, the rpc call from client will only use about 30ms, as our expected.

### DIFF
--- a/proto/table_marshal.go
+++ b/proto/table_marshal.go
@@ -96,7 +96,7 @@ type marshalElemInfo struct {
 
 var (
 	marshalInfoMap  = map[reflect.Type]*marshalInfo{}
-	marshalInfoLock sync.Mutex
+	marshalInfoLock sync.RWMutex
 
 	uint8SliceType = reflect.TypeOf(([]uint8)(nil)).Kind()
 )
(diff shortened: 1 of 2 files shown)


TennyZhuang:
A simple reproduce code can be found at https://github.com/TennyZhuang/protobuf-lock-reproduce

PR #659, opened 2020-01-14
Casttypewith

trasc:
Add a new extension "casttypewith" that could be used to define and use fields
of an non-local type, by using a custom caster type.

The caster should implement:
func (c *Caster) Equal(a, b *Castee) bool
func (c *Caster) Size(a *Castee) int
func (c *Caster) MarshalTo(a *Castee, buf []byte) (int, error)
func (c *Caster) Unmarshal(buf []byte) (*Castee, error)
func (c *Caster) NewPopulated() *Castee

The protobuf field type should be "bytes"
The feature is usable only with marshal and unmarshal plugins
and features like JSONPB and text marshaling is are incompatible.


plugin/castwithfixup/castwithfixup.go
```diff
@@ -0,0 +1,106 @@
...
+// LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+// A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+// OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+// SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
```
AdoAdoAdo: can you add a comment to explain what the plugin does ?

PR #688, opened 2020-05-17
add plugin annotation

u2takey:
this is a proposal and implementation for a new plugin `annotation`, which add an string annotation for message type, which can be used like JAVA's runtime annotation.

Usage:

```
message X {
    option (gogoproto.annotation) = 'a string';
}

message Y {
    // annotation could be a json string
    option (gogoproto.annotation) = '{"A":"a"}';
}
```

Retrieve
```
fmt.Println(gogoproto.GetAnnotations())
fmt.Println(gogoproto.GetAnnotation(&proto.A{}))
```


u2takey:
@awalterschulze @jmarais

PR #700, opened 2020-07-23
Fixed order of imports, make stable generation result

zifter:
It's connected with https://github.com/mwitkow/go-proto-validators/issues/62
Iteration order for map keys is not determinated
https://stackoverflow.com/questions/9619479/go-what-determines-the-iteration-order-for-map-keys


zifter:
I suppose that something wrong with CI for go 1.11.x and vet plugin.

zifter:
@awalterschulze can we merge this pr?

awalterschulze:
Yeah seems legit, thanks for making it easy.

PR #717, opened 2021-01-07
update to go version 1.15.6 and protoc 3.14.0

awalterschulze:
Seems go 1.15.6 was not compiling, this requires a fix in text_parser.go

I also updated the go and protoc versions for github actions and the badge, so hopefully we can get a green build again

PR #721, opened 2021-01-27
Pin versions for dependencies in Makefile

nikhita:
The indirect dependencies in `go.mod`, specifically  `golang.org/x/tools` and `github.com/kisielk/errcheck`, are not really used anywhere in the code. `go mod tidy` also removes these lines. They are only used in the `Makefile` using `go get`.

Currently, no version is specified while using `go get`. This means that go will try to fetch the latest version and this version can deviate from the version in `go.mod`. To avoid this, the PR updates the Makefile to explicitly `go get` the versions specified in `go.mod`.

Additionally, this PR also updates the versions to match the versions in `v1.3.1` i.e. avoids the version bumps made in https://github.com/gogo/protobuf/pull/717:
- `github.com/kisielk/errcheck` to `v1.2.0`
- `golang.org/x/tools` to [`release-branch.go1.15`](https://github.com/golang/tools/tree/release-branch.go1.15)

Context - bumping gogo/protobuf from v1.3.1 to v1.3.2 in kubernetes requires bumping x/tools as well, which we'd like to avoid in k8s (xref https://github.com/kubernetes/client-go/issues/927)

cc @awalterschulze

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,12 +149,12 @@ tests:
 	(cd test/stdtypes && make test)
 
 vet:
-	go get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
+	GO111MODULE=on go get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow@release-branch.go1.15
 	go vet ./...
 	go vet -vettool=$(shell which shadow) ./...
 
 errcheck:
-	go get github.com/kisielk/errcheck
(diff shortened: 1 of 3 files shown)


nikhita:
fyi @dims @palnabarun

awalterschulze:
I guess because it is used in CI and I am not familiar with godeps, it is after my time in go land.
Also see https://github.com/gogo/protobuf/issues/691
Unfortunately we do not have personal capacity to keep on maintaining this project.

PR #740, opened 2021-11-01
add a service regsitry when generate source code

gzzyj:
A lazy and common way for  gRPC service  registration. I can use the registry to invoke the grpc.Server's  function (`func (s *Server) RegisterService(sd *ServiceDesc, ss interface{})` ) at the special time. It is more convenient to manage the registration than directly to invoke the `pkgName.RegisterXXXServer( s *grpc.Server ,svr XXXServer)`


gzzyj (commented):
1. add a serviceoption
2. handle the serviceoption in code gen plugin named grpc。export the serviceDescription

PR #757, opened 2022-09-16
chore: fix typo in equals.go docs

Faagerholm:
typo in equals.go file 

`iff`-> `if` 

Super minor change but I have to look at it every day 😃


bool3max:
Not a typo, "iff" is intentional ("if and only if")

PR #779, opened 2026-03-13
proto: optimize Properties.Parse() using IndexByte

AmritM18:
Replace strings.Split with strings.IndexByte for better performance:
- Eliminates slice allocation for field parsing
- Single-pass parsing without intermediate slices

### DIFF
--- a/proto/properties.go
+++ b/proto/properties.go
@@ -202,13 +202,13 @@ func (p *Properties) String() string {
 // Parse populates p by parsing a string in the protobuf struct field tag style.
 func (p *Properties) Parse(s string) {
 	// "bytes,49,opt,name=foo,def=hello!"
-	fields := strings.Split(s, ",") // breaks def=, but handled below.
-	if len(fields) < 2 {
+	comma := strings.IndexByte(s, ',')
+	if comma == -1 {
 		log.Printf("proto: tag has too few fields: %q", s)
 		return
 	}
 
-	p.Wire = fields[0]
+	p.Wire = s[:comma]
 	switch p.Wire {
 	case "varint":
 		p.WireType = WireVarint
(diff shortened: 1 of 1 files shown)


proto/properties.go
```diff
@@ -259,17 +279,32 @@ outer:
...
 			}
+			break outer
 		case strings.HasPrefix(f, "embedded="):
-			p.OrigName = strings.Split(f, "=")[1]
+			value := f[9:]
+			if eq := strings.IndexByte(value, '='); eq != -1 {
+				p.OrigName = value[:eq]
+			} else {
+				p.OrigName = value
+			}
```
rabbbit: should we just do https://pkg.go.dev/strings#CutPrefix here and below?

it would be nicer if we could do this as part of the switch, but I think that's a no?

proto/properties.go
```diff
@@ -228,15 +228,35 @@ func (p *Properties) Parse(s string) {
 		return
 	}
 
+	pos := comma + 1
+	comma = strings.IndexByte(s[pos:], ',')
+	var tagField string
+	if comma == -1 {
+		tagField = s[pos:]
+		pos = len(s)
```
rabbbit, line 236: why is this here, shouldn't this be an error?

(didn't previous code reqire at least 2 commas?(

AmritM18: hm I somehow missed this. yeah this code isn't super easy to read hm. I can try out the splitN optimization.

as for this code, we're looking at the second comma here so we can get the tag field. the original code checked if there were at least 2 items after splitting by commas. the check on line 206 accomplishes the same thing (2 items == 1 comma).

proto/properties.go
AmritM18: yah sure, I can switch for that. I don't know how it'd be part of the switch though.

proto/properties.go
```diff
@@ -247,29 +267,44 @@ outer:
...
 				// Commas aren't escaped, and def is always last.
-				p.Default += "," + strings.Join(fields[i+1:], ",")
-				break outer
+				p.Default += "," + s[pos:]
 			}
+			break outer
 		case strings.HasPrefix(f, "embedded="):
-			p.OrigName = strings.Split(f, "=")[1]
+			value, _ := strings.CutPrefix(f, "embedded=")
+			if eq := strings.IndexByte(value, '='); eq != -1 {
```
rabbbit, line 289: why do we need all this indexing? shouldn't but prefix be enough?

